PR #18, opened 2024-04-18
add pam-bw conveter

the-infinity:
Adds "Baden-Württemberg: Park und Mitfahren" converter


src/parkapi_sources/converters/pam_bw/converter.py
```diff
@@ -0,0 +1,77 @@
...
+from datetime import datetime, timezone
+from typing import Any
+
+from openpyxl.cell import Cell
+from openpyxl.workbook.workbook import Workbook
+from validataclass.exceptions import ValidationError
+from validataclass.validators import DataclassValidator
+
+from parkapi_sources.converters.base_converter.push import XlsxConverter
+from parkapi_sources.converters.neckarsulm.models import NeckarsulmRowInput
```
AbdullahiFatola, line 15: Do we still need Neckarsulm Model?

src/parkapi_sources/converters/pam_bw/converter.py
```diff
@@ -0,0 +1,77 @@
...
+from parkapi_sources.exceptions import ImportParkingSiteException
+from parkapi_sources.models import SourceInfo, StaticParkingSiteInput
+
+
+class PamBwPushConverter(XlsxConverter):
+    neckarsulm_row_validator = DataclassValidator(NeckarsulmRowInput)
```
AbdullahiFatola, line 21: Neckarsulm row validator is also here. I guess its a mistake?

derhuerst (commented):
Can you explain what "PAM" means, e.g. in a comment on the converter's class?

the-infinity:
> Can you explain what "PAM" means, e.g. in a comment on the converter's class?

German-english mixup (a = and). Changed it to German name and added it to the README.

src/parkapi_sources/converters/pam_bw/converter.py
the-infinity: Nope, copy paste leftover. Removed.

src/parkapi_sources/converters/pam_bw/converter.py
the-infinity: Yep, copy paste leftover. Removed.

PR #26, opened 2024-04-30
Ellwangen Push Converter

AbdullahiFatola:
This PR integrates the parking data for cars in the city of Ellwangen


src/parkapi_sources/converters/ellwangen/converter.py
```diff
@@ -0,0 +1,66 @@
...
+class EllwangenPushConverter(NormalizedXlsxConverter):
+
+    source_info = SourceInfo(
+        uid='ellwangen',
+        name='Ellwangen Parkdaten',
+        public_url='https://www.ellwangen.de',
+        has_realtime_data=False,
+    )
+
+    header_row: dict[str, str] = {
```
the-infinity, line 24: Why do you redefine the whole `header_row` when just one line has changed?

src/parkapi_sources/converters/ellwangen/converter.py
```diff
@@ -0,0 +1,66 @@
...
+        'ID': 'uid',
+        'Name': 'name',
+        'Art der Anlage': 'type',
+        'Betreiber Name': 'operator_name',
+        'Längengrad': 'lat',
+        'Breitengrad': 'lon',
+        'Adresse mit PLZ und Stadt': 'address',
+        'Maximale Parkdauer / min': 'max_stay',
```
the-infinity, line 32: Is `/ min` part of the template now? Then I would suggest to changed it in `NormalizedXlsxConverter`.

README.md
```diff
@@ -13,6 +13,7 @@ We support following data sources:
 | Barrierefreie Reisekette Baden-Württemberg: PKW-Parkplätze an Bushaltestellen     | car     | push (csv)  | `bfrk_bw_spnv_car`   | no       |
 | Barrierefreie Reisekette Baden-Württemberg: Fahrrad-Parkplätze an Bahnhöfen       | bike    | push (csv)  | `bfrk_bw_oepnv_bike` | no       |
 | Barrierefreie Reisekette Baden-Württemberg: Fahrrad-Parkplätze an Bushaltestellen | bike    | push (csv)  | `bfrk_bw_spnv_bike`  | no       |
+| Stadt Ellwangen                                                                   | car     | push (excel)| `ellwangen`          | no       |
```
the-infinity, line 16: As Excel supports different formats, I would use the file ending here (`xlsx`).

AbdullahiFatola: Yeah that makes sense!

src/parkapi_sources/converters/ellwangen/converter.py
the-infinity: Hmm. To be clear about that: the normalized converter was meant as an template matching the template we provided at the start. Idea of that template was that it is not changed by cities. If you accept changed templates, this means no generic converter, but an adaption to every city who changes something. Would not recommend, but that's up to you :)

src/parkapi_sources/converters/ellwangen/converter.py
the-infinity: If that's the new template this is fine.

the-infinity:
Your ruff linting failes atm, please fix this.

AbdullahiFatola:
> Your ruff linting failes atm, please fix this.

AbdullahiFatola:
> Your ruff linting failes atm, please fix this.

It's now fixed it!

src/parkapi_sources/converters/ellwangen/converter.py
AbdullahiFatola: Is it possible to define the ```header_row``` like this ```header_row: dict[str, str] = { 'Maximale Parkdauer / min': 'max_stay'}```. Such that, all other attributes from the ellwangen converter will be inherited from the ```header_row``` in the ```NormalizedXlsxConverter```. This way, i will not need to modify or change anything the ```header_row``` in the ```NormalizedXlsxConverter```

I tried it but i get the error ```AssertionError: There should be more valid then invalid parking sites```
Would that be possible to implement?

the-infinity: `AssertionError: There should be more valid then invalid parking sites` - then you basically broke the whole converter, because that's actually what this test is for: to test of the converter gives a reasonable amount of results. So, this is by purpose, to prevent that any changes breaks the converter.

You can re-use a large dictionary from a parent class by extending it in your constructor. Then you don't need to re-define everything. Eg:

```
def __init__(...):
  self.header_row = {**super().header_row, 'new_key': 'new_value'}
```

src/parkapi_sources/converters/ellwangen/converter.py
```diff
@@ -0,0 +1,36 @@
...
+        has_realtime_data=False,
+    )
+
+    def map_row_to_parking_site_dict(self, mapping: dict[str, int], row: list[Cell]) -> dict[str, Any]:
+        parking_site_dict: dict[str, str] = {}
+
+        for field in mapping.keys():
+            parking_site_dict[field] = row[mapping[field]].value
+
+        # TO DO: Maximale Parkdauer is in Minutes, which should be converted to seconds
```
the-infinity, line 30: Why is this a TODO?

src/parkapi_sources/converters/ellwangen/converter.py
```diff
@@ -0,0 +1,36 @@
...
+        parking_site_dict['max_stay'] = parking_site_dict['max_stay'] * 60 if isinstance(parking_site_dict['max_stay'], int) else None
```
the-infinity, line 31: Why don't you just check for `None`?

src/parkapi_sources/converters/base_converter/push/normalized_xlsx_converter.py
```diff
@@ -62,6 +72,23 @@ class NormalizedXlsxConverter(XlsxConverter, ABC):
         'Am Straßenrand': 'ON_STREET',
     }
 
+    def get_mapping_by_header(self, row: tuple[Cell]) -> dict[str, int]:
```
the-infinity, line 75: Why did you introduce `row_values_no_special_chars`? Don't see any reason for this, as `{'some.count': 'capacity'}` is a perfectly valid dictionary which will work for the mapping.

src/parkapi_sources/converters/ellwangen/converter.py
AbdullahiFatola: To simplify the converter, i have redefined the ```header_row``` as whole

src/parkapi_sources/converters/ellwangen/converter.py
AbdullahiFatola: Okay I would just redefined the ```header_row``` as a whole

src/parkapi_sources/converters/base_converter/push/normalized_xlsx_converter.py
AbdullahiFatola: I would just redefined the ```header_row``` as a whole to avoid complications.

PR #52, opened 2024-06-07
Apcoa API pull converter

AbdullahiFatola:
This PR integrates the pull converter for the static parking data pulled from the APCOA-SERVICES API.


src/parkapi_sources/converters/apcoa/converter.py
```diff
@@ -0,0 +1,70 @@
...
+
+    source_info = SourceInfo(
+        uid='apcoa',
+        name='APCOA-SERVICES API',
+        public_url='https://devzone.apcoa-services.com/',
+        has_realtime_data=False,  # ATM only static data can be called from the API
+    )
+
+    def get_static_parking_sites(self) -> tuple[list[StaticParkingSiteInput], list[ImportParkingSiteException]]:
+        static_parking_site_inputs: list = []
```
the-infinity, line 33: Missing typing.

src/parkapi_sources/converters/apcoa/converter.py
```diff
@@ -0,0 +1,70 @@
...
+        static_parking_site_errors: list[ImportParkingSiteException] = []
+
+        parking_site_dicts = self.get_data()
+
+        for parking_site_dict in parking_site_dicts.get('Results', []):
```
the-infinity, line 38: Missing validation: what happens, if there is no dict? -> ListValidator with AnythingValidator enforcing dicts inside the list.

src/parkapi_sources/converters/apcoa/converter.py
```diff
@@ -0,0 +1,70 @@
...
+        return [], []  # ATM only static data can be called from the API
+
+    def get_data(self) -> dict:
+        headers: dict[str, str] = {
+            'Cache-Control': 'no-cache',
+            'Ocp-Apim-Subscription-Key': self.config_helper.get('PARK_API_APCOA_API_SUBSCRIPTION_KEY'),
+        }
+
+        response = requests.get(
+            f'{self.config_helper.get("PARK_API_APCOA_URL", self._base_url)}/Carparks',
```
the-infinity, line 66: Why load this from config, and why not using `source_info.source_url`?

src/parkapi_sources/converters/apcoa/validators.py
```diff
@@ -0,0 +1,125 @@
...
+            self.OPEN_SURFACE: ParkingSiteType.OFF_STREET_PARKING_GROUND,
+        }.get(self, ParkingSiteType.OTHER)
+
+
+class ApcoaNavigationLocationType:
+    DEFAULT = 'default'
+    CAR_ENTRY = 'CarEntry'
+
+
+class ApcoaOpeningHoursWeekdays:
```
the-infinity, line 64: This is not a validator.

src/parkapi_sources/converters/apcoa/mapper.py
```diff
@@ -0,0 +1,83 @@
...
+                and opening_hours_input.OpeningTimes == ApcoaOpeningHoursWeekdays.OpeningTime
+            ]
+        )
+        if num_of_weekdays == 7:
+            static_parking_site_input.opening_hours = '24/7'
+
+        for capacity_data in apcoa_input.Spaces:
+            # Because it was checked in validation, we can be sure that capacity will be set
+            if capacity_data.Type == ApcoaParkingSpaceType.TOTAL_SPACES:
+                static_parking_site_input.capacity = int(round(capacity_data.Count))
```
the-infinity, line 59: Why do you do `int(round())`? `capacity_data.Count` is an int, you know this because you validated it before.

src/parkapi_sources/converters/apcoa/mapper.py
```diff
@@ -0,0 +1,83 @@
...
+            elif capacity_data.Type == ApcoaParkingSpaceType.DISABLED_SPACES:
+                static_parking_site_input.capacity_disabled = int(round(capacity_data.Count))
+            elif capacity_data.Type == ApcoaParkingSpaceType.WOMEN_SPACES:
+                static_parking_site_input.capacity_woman = int(round(capacity_data.Count))
+            elif capacity_data.Type == ApcoaParkingSpaceType.ELECTRIC_CAR_CHARGING_SPACES:
```
the-infinity, line 64: Why not using `capacity_data.Type in [ApcoaParkingSpaceType.ENUM_1, ApcoaParkingSpaceType.ENUM2]`?

src/parkapi_sources/converters/apcoa/mapper.py
```diff
@@ -0,0 +1,83 @@
...
+                    for location_type_input in apcoa_input.NavigationLocations
+                    if location_type_input.LocationType == ApcoaNavigationLocationType.DEFAULT
+                )
+            ),
+            purpose=PurposeType.CAR,
+            type=apcoa_input.CarparkType.Name.to_parking_site_type_input(),
+            has_realtime_data=False,  # TODO: change this as soon as Apcoa API offers realtime data
+            static_data_updated_at=datetime.now(tz=timezone.utc),
+        )
+        if apcoa_input.CarParkWebsiteURL:
```
the-infinity, line 39: Why this `if`?

src/parkapi_sources/converters/apcoa/mapper.py
```diff
@@ -0,0 +1,83 @@
...
+from .validators import ApcoaNavigationLocationType, ApcoaOpeningHoursWeekdays, ApcoaParkingSiteInput, ApcoaParkingSpaceType
+
+
+class ApcoaMapper:
+    @staticmethod
+    def map_static_parking_site(apcoa_input: ApcoaParkingSiteInput) -> StaticParkingSiteInput:
+        static_parking_site_input = StaticParkingSiteInput(
+            uid=str(apcoa_input.CarParkId),
+            name=apcoa_input.CarparkLongName if apcoa_input.CarparkLongName else apcoa_input.CarparkShortName,
+            lat=next(
```
the-infinity, line 20: 1) You do the same iteration two times. Why not do it just once?
2) When there is no entry with`ApcoaNavigationLocationType.DEFAULT`, you will have an exception for the whole converter which is unwanted

src/parkapi_sources/converters/apcoa/mapper.py
```diff
@@ -0,0 +1,83 @@
...
+                iter(
+                    location_type_input.GeoCoordinates.Longitude
+                    for location_type_input in apcoa_input.NavigationLocations
+                    if location_type_input.LocationType == ApcoaNavigationLocationType.DEFAULT
+                )
+            ),
+            purpose=PurposeType.CAR,
+            type=apcoa_input.CarparkType.Name.to_parking_site_type_input(),
+            has_realtime_data=False,  # TODO: change this as soon as Apcoa API offers realtime data
+            static_data_updated_at=datetime.now(tz=timezone.utc),
```
the-infinity, line 37: But you do have a last updated at `LastModifiedDateTime`

src/parkapi_sources/converters/apcoa/mapper.py
```diff
@@ -0,0 +1,83 @@
...
+
+        num_of_weekdays = len(
+            [
+                opening_hours_input.Weekday
+                for opening_hours_input in apcoa_input.OpeningHours
+                if opening_hours_input.Weekday in ApcoaOpeningHoursWeekdays.Weekdays
+                and opening_hours_input.OpeningTimes == ApcoaOpeningHoursWeekdays.OpeningTime
+            ]
+        )
+        if num_of_weekdays == 7:
```
the-infinity, line 53: As the data is well structured, you could do a better mapping to OSM format to get the other opening times, too.

src/parkapi_sources/converters/apcoa/validators.py
```diff
@@ -0,0 +1,125 @@
...
+
+@validataclass
+class ApcoaParkingSpaceInput:
+    Type: str = StringValidator()
+    Count: int = IntegerValidator(allow_strings=True, min_value=0)
+
+
+@validataclass
+class ApcoaOpeningHoursWeekdayInput:
+    Weekday: str = StringValidator()
```
the-infinity, line 102: This should be an enumeration

src/parkapi_sources/converters/apcoa/validators.py
```diff
@@ -0,0 +1,125 @@
...
+    CarparkLongName: Optional[str] = Noneable(StringValidator())
+    CarparkShortName: Optional[str] = Noneable(StringValidator())
+    CarParkWebsiteURL: Optional[str] = Noneable(UrlValidator())
+    CarparkType: ApcoaCarparkTypeNameInput = DataclassValidator(ApcoaCarparkTypeNameInput)
+    Address: ApcoaAdressInput = DataclassValidator(ApcoaAdressInput)
+    NavigationLocations: list[ApcoaNavigationLocationsInput] = ListValidator(DataclassValidator(ApcoaNavigationLocationsInput))
+    Spaces: list[ApcoaParkingSpaceInput] = ListValidator(DataclassValidator(ApcoaParkingSpaceInput))
+    OpeningHours: list[ApcoaOpeningHoursWeekdayInput] = ListValidator(DataclassValidator(ApcoaOpeningHoursWeekdayInput))
+
+    # TODO: ignored multiple attributes which do not matter so far
```
the-infinity, line 118: As we support `photo_url`, you could add `CarParkPhotoURLs` to your validation and mapping.

src/parkapi_sources/converters/apcoa/validators.py
AbdullahiFatola: Now modified!

src/parkapi_sources/converters/apcoa/mapper.py
AbdullahiFatola: Could you please elaborate on what the values of other OSM-Formats should be? i have added some but not sure if they are right.

src/parkapi_sources/converters/apcoa/mapper.py
AbdullahiFatola: That's right, now added.

src/parkapi_sources/converters/apcoa/mapper.py
AbdullahiFatola: Oh that's right, i have now removed the second iteration and checked for no entry in LocationType

src/parkapi_sources/converters/apcoa/mapper.py
AbdullahiFatola: To check if ```CarParkWebsiteURL``` has no entry.

src/parkapi_sources/converters/apcoa/mapper.py
AbdullahiFatola: Oh yeah, that's right.

src/parkapi_sources/converters/apcoa/validators.py
AbdullahiFatola: It's now made an enum validator

src/parkapi_sources/converters/apcoa/converter.py
AbdullahiFatola: now changed!

src/parkapi_sources/converters/apcoa/converter.py
AbdullahiFatola: That right, it is now updated.

src/parkapi_sources/converters/apcoa/validators.py
the-infinity: If a value is null, you can use `Noneable` for accepting `None`: `foo: Optional[str] = Noneable(StringValidator())`

src/parkapi_sources/converters/apcoa/mapper.py
the-infinity: Voila: https://wiki.openstreetmap.org/wiki/Key:opening_hours

src/parkapi_sources/converters/apcoa/mapper.py
the-infinity: Why `apcoa_input.LastModifiedDateTime if apcoa_input.LastModifiedDateTime else datetime.now(tz=timezone.utc),`? In which condition `datetime.now(tz=timezone.utc)` will be used?

src/parkapi_sources/converters/apcoa/validators.py
```diff
@@ -0,0 +1,162 @@
...
+class ApcoaOpeningHoursWeekdays(Enum):
+    MONDAY = 'Monday'
+    TUESDAY = 'Tuesday'
+    WEDNESDAY = 'Wednesday'
+    THURSDAY = 'Thursday'
+    FRIDAY = 'Friday'
+    SATURDAY = 'Saturday'
+    SUNDAY = 'Sunday'
+
+    def to_osm_opening_day_format(self) -> Enum:
```
the-infinity, line 76: This does not return an Enum

src/parkapi_sources/converters/apcoa/validators.py
```diff
@@ -0,0 +1,162 @@
...
+            self.TUESDAY: 'Tu',
+            self.WEDNESDAY: 'We',
+            self.THURSDAY: 'Th',
+            self.FRIDAY: 'Fr',
+            self.SATURDAY: 'Sa',
+            self.SUNDAY: 'Su',
+        }.get(self, None)
+
+
+class ApcoaOpeningHoursTime(Enum):
```
the-infinity, line 88: Why do you have an Enum here? What should be enumerated?

src/parkapi_sources/converters/apcoa/mapper.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+class ApcoaMapper:
+    @staticmethod
+    def map_static_parking_site(apcoa_input: ApcoaParkingSiteInput) -> StaticParkingSiteInput:
+        latitude, longitude = next(
+            iter(
+                (navigation_locations_input.GeoCoordinates.Latitude, navigation_locations_input.GeoCoordinates.Longitude)
+                if navigation_locations_input.LocationType == ApcoaNavigationLocationType.DEFAULT
+                else (None, None)
```
the-infinity, line 28: Why this line? You This gives invalid data in case of missing `ApcoaNavigationLocationType.DEFAULT`. You should check for the existence of `ApcoaNavigationLocationType.DEFAULT`  in your validator, and here just rely on that.

the-infinity:
Registration in main file and README is missing

src/parkapi_sources/converters/apcoa/converter.py
the-infinity: Why appending `/Carparks` here?

src/parkapi_sources/converters/apcoa/validators.py
AbdullahiFatola: Ok amended

src/parkapi_sources/converters/apcoa/mapper.py
AbdullahiFatola: In the case of no entries for ```static_data_updated_at``` to avoid null values.

src/parkapi_sources/converters/apcoa/converter.py
AbdullahiFatola: I assumed the ```source_url``` is the main endpoint and for flexibility, i defined only the main endpoint in ```source_url```should the converter need to get realtime data from other endpoints of the API later in the future. I have however removed it for now.

AbdullahiFatola:
> Registration in main file and README is missing

Thanks for reminding

src/parkapi_sources/converters/apcoa/converter.py
the-infinity: In theory correct thought, in practice: do you know any other information which is relevant for ParkAPI which is provided there?

src/parkapi_sources/converters/apcoa/mapper.py
the-infinity: But you validated `LastModifiedDateTime` before so that it will never become `None`:

``` 
    LastModifiedDateTime: datetime = SpacedDateTimeValidator(
        local_timezone=ZoneInfo('Europe/Berlin'),
        target_timezone=timezone.utc,
    )
```

src/parkapi_sources/converters/apcoa/mapper.py
```diff
@@ -0,0 +1,101 @@
...
+                static_parking_site_input.opening_hours = '24/7'
+            elif (
+                len(weekdays) == 5
+                and ApcoaOpeningHoursWeekdays.SATURDAY not in weekdays
+                and ApcoaOpeningHoursWeekdays.SUNDAY not in weekdays
+            ):
+                static_parking_site_input.opening_hours = f'Mo-Fr {opening_time}'
+            else:
+                static_parking_site_input.opening_hours = ';'.join([f'{weekday} {opening_time}' for weekday in weekdays])
+
```
the-infinity, line 74: After I really need some concentration to understand what happens in OSM transformation: this would be a good part of code to move to a method just for that purpose, because it's far from trivial, and it would make sense to make a unittest with parameters to test different inputs. At our validators, we have some examples for tests of multiple inputs: https://github.com/ParkenDD/parkapi-sources-v3/blob/main/tests/validators/integer_validators_test.py#L14

src/parkapi_sources/converters/apcoa/converter.py
AbdullahiFatola: I just added tariff information, which i mapped to ```has_fee```. Aside this, I don't think there are other information for now.
**Short question**: there is a ```CarparkType``` with value of Multi Level Car Park (```MLCP```), which I mapped to ```OFF_STREET_PARKING_GROUND```. Is this correct?

src/parkapi_sources/converters/apcoa/mapper.py
AbdullahiFatola: That's right.

src/parkapi_sources/converters/apcoa/validators.py
AbdullahiFatola: it's now used as string directly

src/parkapi_sources/converters/apcoa/validators.py
```diff
@@ -0,0 +1,161 @@
...
+
+@validataclass
+class ApcoaOpeningHoursInput:
+    Weekday: ApcoaOpeningHoursWeekdays = EnumValidator(ApcoaOpeningHoursWeekdays)
+    OpeningTimes: str = StringValidator()
+
+
+@validataclass
+class ApcoaCarparkPhotoURLInput:
+    CarparkPhotoURL1: Optional[str] = Noneable(StringValidator())
```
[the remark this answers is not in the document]
AbdullahiFatola: Good tip, thanks

src/parkapi_sources/converters/apcoa/mapper.py
```diff
@@ -0,0 +1,101 @@
...
+            static_data_updated_at=apcoa_input.LastModifiedDateTime,
+        )
+
+        if apcoa_input.Address.Street and apcoa_input.Address.Zip and apcoa_input.Address.City:
+            static_parking_site_input.address = f'{apcoa_input.Address.Street}, {apcoa_input.Address.Zip} {apcoa_input.Address.City}'
+
+        if apcoa_input.CarParkPhotoURLs:
+            static_parking_site_input.photo_url = apcoa_input.CarParkPhotoURLs.CarparkPhotoURL1
+
+        apcoa_weekdays = [
```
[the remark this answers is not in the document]
AbdullahiFatola: nice tips :)

src/parkapi_sources/converters/apcoa/mapper.py
```diff
@@ -0,0 +1,101 @@
...
+            ApcoaOpeningHoursWeekdays.MONDAY,
+            ApcoaOpeningHoursWeekdays.TUESDAY,
+            ApcoaOpeningHoursWeekdays.WEDNESDAY,
+            ApcoaOpeningHoursWeekdays.THURSDAY,
+            ApcoaOpeningHoursWeekdays.FRIDAY,
+            ApcoaOpeningHoursWeekdays.SATURDAY,
+            ApcoaOpeningHoursWeekdays.SUNDAY,
+        ]
+        apcoa_opening_times = defaultdict(list)
```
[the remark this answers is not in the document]
AbdullahiFatola: Now added

src/parkapi_sources/converters/apcoa/mapper.py
AbdullahiFatola: It is now made a method for the purpose. Is it okay in the validator class?

the-infinity: Yes, the approach is good, but this is really something which makes sense to create unittests, as there are a lot of different inputs.

src/parkapi_sources/converters/apcoa/validators.py
```diff
@@ -0,0 +1,194 @@
...
+        raise ValidationError(reason='Missing parking spaces capacity')
+
+    def get_osm_opening_hours(self) -> str:
+        apcoa_opening_times: dict = defaultdict(list)
+        for opening_hours_input in self.OpeningHours:
+            opening_time = opening_hours_input.OpeningTimes.replace(' ', '')
+            if opening_hours_input.OpeningTimes == '00:00 - 00:00':
+                apcoa_opening_times['24/7'].append(opening_time)
+            if opening_hours_input.Weekday in list(ApcoaOpeningHoursWeekdays)[:5]:
+                apcoa_opening_times['Mo-Fr'].append(opening_time)
```
the-infinity, line 181: This is quite a good example why you need unittests for this transformation. When you have an input like `Monday: 10:00 - 20:00, Tuesday: 10:00 - 12:00, Saturday: 10:00 - 16:00`, this ends up in `Mo-Fr: 10:00 - 20:00, 10:00 - 12:00; Sa: 10:00 - 16:00`.

src/parkapi_sources/converters/apcoa/mapper.py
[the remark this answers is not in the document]
the-infinity: `dict[str, list[str]]` would be more specific.

src/parkapi_sources/converters/apcoa/validators.py
the-infinity: Added an example unittest testing common transformations

src/parkapi_sources/converters/apcoa/converter.py
the-infinity: It's more `CAR_PARK`, because `OFF_STREET_PARKING_GROUND` is just a flat ground.

src/parkapi_sources/converters/apcoa/validators.py
```diff
@@ -0,0 +1,224 @@
...
+            if opening_hours_input.Weekday in list(ApcoaOpeningHoursWeekday)[:5] and opening_time != 'closed':
+                apcoa_opening_times['Mo-Fr'].append(opening_time)
+            if opening_hours_input.Weekday in list(ApcoaOpeningHoursWeekday) and opening_time != 'closed':
+                apcoa_opening_times[opening_hours_input.Weekday.value].append(opening_time)
+
+        osm_opening_hour: list = []
+        if len(apcoa_opening_times['Mo-Fr']) == 5 and len(set(apcoa_opening_times['Mo-Fr'])) == 1:
+            osm_opening_hour.append(f'Mo-Fr {next(iter(set(apcoa_opening_times["Mo-Fr"])))}')
+        else:
+            if 'Monday' in list(apcoa_opening_times.keys()):
```
the-infinity, line 190: Why not iterating over list(ApcoaOpeningHoursWeekday)?

``` 
for weekday in list(ApcoaOpeningHoursWeekday)[:5]:
  print(weekday.name) -> MONDAY
  print(weekday.value) -> Monday
  print(weekday.to_osm_opening_day_format()) -> Mo
```

AbdullahiFatola: good tip, thanks!

the-infinity:
Added some additional comments and improvements here and merged it: https://github.com/ParkenDD/parkapi-sources-v3/pull/72

PR #70, opened 2024-07-11
Opendata Swiss pull converter

AbdullahiFatola:
The PR integrates the pull converter for the Park & Rail/Park & Ride Static Data from the Open Data Portal of Switzerland, also known as Open Data Swiss or opentransport.swiss


src/parkapi_sources/converters/opendata_swiss/converter.py
```diff
@@ -0,0 +1,79 @@
...
+from validataclass.validators import DataclassValidator
+
+from parkapi_sources.converters.base_converter.pull import GeojsonInput, PullConverter
+from parkapi_sources.exceptions import ImportParkingSiteException, ImportSourceException
+from parkapi_sources.models import RealtimeParkingSiteInput, SourceInfo, StaticParkingSiteInput
+
+from .models import OpenDataSwissFeatureInput
+
+
+class OpenDataSwissBasePullConverter(PullConverter, ABC):
```
the-infinity, line 19: Why do you create a Base class for just one Child class? Do we expect more Child classes?

src/parkapi_sources/converters/opendata_swiss/converter.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+class OpenDataSwissPullConverter(OpenDataSwissBasePullConverter):
+    opendata_swiss_feature_validator = DataclassValidator(OpenDataSwissFeatureInput)
+
+    source_info = SourceInfo(
+        uid='opendata_swiss',
+        name='Open-Data-Plattform öV Schweiz (opentransport.swiss)',
+        public_url='https://opentransportdata.swiss/de/dataset/parking-facilities',
+        source_url='https://opentransportdata.swiss/dataset/c1be030e-adad-41c6-8ee2-899e85f840f6/resource/5dab3aba-6b5a-42cb-bb73-103fb28c7e10/download/parking-facilities.json',
```
the-infinity, line 72: This `source_url` ends up in an HTTP 500. If I understand it correctly, the permalink is `https://opentransportdata.swiss/de/dataset/parking-facilities/permalink` instead, and you have to make sure that you follow the HTTP 302.

src/parkapi_sources/converters/opendata_swiss/models.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+@validataclass
+class OpenDataSwissAddressInput:
+    addressLine: Optional[str] = Noneable(StringValidator())
+    city: Optional[str] = Noneable(StringValidator())
+    postalCode: Optional[str] = Noneable(StringValidator())
+
+
+class OpenDataSwissCapacityCategoryTypeInput(Enum):
```
the-infinity, line 27: We usually use `Input` for `validataclasses`, not for Enums. I would suggest to call it `OpenDataSwissCapacityCategoryType`, or even `OpenDataSwissCapacityCategory`, because `Category` is already quite clear about what that is.

src/parkapi_sources/converters/opendata_swiss/models.py
```diff
@@ -0,0 +1,200 @@
...
+        opening_hh_mm_ss: list[str] = list(self.operationTime.operatingFrom.split(':'))
+        opening_time: str = f'{opening_hh_mm_ss[0]}:{opening_hh_mm_ss[1]}-{closing_hh_mm_ss[0]}:{closing_hh_mm_ss[1]}'
+
+        for days_of_week_input in self.operationTime.daysOfWeek:
+            # If it's open all day, add it to our 24/7 check counter, and we change opening_time to the OSM format
+            if opening_time == '00:00-24:00':
+                check_counter_24_7 += 1
+
+            # Add opening times to fallback dict
+            open_swiss_opening_times_by_weekday[days_of_week_input].append(opening_time)
```
the-infinity, line 113: I'm not sure if that makes much sense. If I understand the format correctly, the times are always the same, so why listing them at all? In general, if I understand the format correctly, you can simplify this method a lot.

src/parkapi_sources/converters/opendata_swiss/models.py
```diff
@@ -0,0 +1,200 @@
...
+            static_data_updated_at=datetime.now(tz=timezone.utc),
+            has_realtime_data=False,
+        )
+
+        if self.properties.parkingFacilityCategory == PurposeType.CAR:
+            static_parking_site_input.purpose = PurposeType.CAR
+            static_parking_site_input.type = ParkingSiteType.CAR_PARK
+        elif self.properties.parkingFacilityCategory == PurposeType.BIKE:
+            static_parking_site_input.purpose = PurposeType.BIKE
+            static_parking_site_input.type = ParkingSiteType.OFF_STREET_PARKING_GROUND
```
the-infinity, line 166: `OFF_STREET_PARKING_GROUND` is for cars as the enum documents. There is a `BIKE_GENERIC` for unknown bike types instead.

src/parkapi_sources/converters/opendata_swiss/models.py
```diff
@@ -0,0 +1,200 @@
...
+            uid=str(self.id),
+            name=self.properties.displayName,
+            operator_name=self.properties.operator,
+            lat=self.geometry.coordinates[1],
+            lon=self.geometry.coordinates[0],
+            static_data_updated_at=datetime.now(tz=timezone.utc),
+            has_realtime_data=False,
+        )
+
+        if self.properties.parkingFacilityCategory == PurposeType.CAR:
```
the-infinity, line 161: This and the `elif` will never be True, as `self.properties.parkingFacilityCategory` is a string, not `PurposeType`.

src/parkapi_sources/converters/opendata_swiss/models.py
```diff
@@ -0,0 +1,200 @@
...
+
+@validataclass
+class OpenDataSwissCapacitiesInput:
+    categoryType: OpenDataSwissCapacityCategoryTypeInput = EnumValidator(OpenDataSwissCapacityCategoryTypeInput)
+    total: int = IntegerValidator()
+
+
+@validataclass
+class OpenDataSwissAdditionalInformationInput:
+    de: Optional[str] = Noneable(StringValidator(multiline=False, unsafe=True))
```
the-infinity, line 42: Please make sure that unsafe strings are handled later, so that they are safe in the end (most likely with an replacing regexp). Otherwise, you will end up in validation errors if you integrate these strings in ParkAPI.

src/parkapi_sources/converters/opendata_swiss/models.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def get_osm_opening_hours(self) -> str:
+        open_swiss_opening_times_by_weekday: dict[OpenDataSwissOperationTimeDaysOfWeek, list[str]] = defaultdict(list)
+        check_counter_24_7: int = 0
+        check_list_weekday: list[str] = []
+
+        # OSM 24/7 has no secs in its timeformat and no endtime 00:00, so we replace with 24:00 and remove the secs
+        closing_hh_mm_ss: list[str] = list(self.operationTime.operatingTo.replace('00:00:00', '24:00:00').split(':'))
+        opening_hh_mm_ss: list[str] = list(self.operationTime.operatingFrom.split(':'))
+        opening_time: str = f'{opening_hh_mm_ss[0]}:{opening_hh_mm_ss[1]}-{closing_hh_mm_ss[0]}:{closing_hh_mm_ss[1]}'
```
the-infinity, line 105: You do not validate the time format before, so `opening_hh_mm_ss[1]` can run into an exception. I would suggest to use a `TimeValidator` in order to catch these issues right at the start.

src/parkapi_sources/converters/opendata_swiss/models.py
```diff
@@ -0,0 +1,200 @@
...
+        # If no capacity with type PARKING was found, we miss the capacity and therefore throw a validation error
+        raise ValidationError(reason='Missing parking spaces capacity')
+
+    def get_osm_opening_hours(self) -> str:
+        open_swiss_opening_times_by_weekday: dict[OpenDataSwissOperationTimeDaysOfWeek, list[str]] = defaultdict(list)
+        check_counter_24_7: int = 0
+        check_list_weekday: list[str] = []
+
+        # OSM 24/7 has no secs in its timeformat and no endtime 00:00, so we replace with 24:00 and remove the secs
+        closing_hh_mm_ss: list[str] = list(self.operationTime.operatingTo.replace('00:00:00', '24:00:00').split(':'))
```
the-infinity, line 103: `self.operationTime.operatingTo` can be None, which would lead into an Exception

src/parkapi_sources/converters/opendata_swiss/models.py
```diff
@@ -0,0 +1,200 @@
...
+        for capacity in self.capacities:
+            if capacity.categoryType == OpenDataSwissCapacityCategoryTypeInput.STANDARD:
+                return
+        # If no capacity with type PARKING was found, we miss the capacity and therefore throw a validation error
+        raise ValidationError(reason='Missing parking spaces capacity')
+
+    def get_osm_opening_hours(self) -> str:
+        open_swiss_opening_times_by_weekday: dict[OpenDataSwissOperationTimeDaysOfWeek, list[str]] = defaultdict(list)
+        check_counter_24_7: int = 0
+        check_list_weekday: list[str] = []
```
the-infinity, line 100: Why collecting these at all, if the times are always the same by format definition?

src/parkapi_sources/converters/opendata_swiss/models.py
```diff
@@ -0,0 +1,200 @@
...
+    def __post_init__(self):
+        for capacity in self.capacities:
+            if capacity.categoryType == OpenDataSwissCapacityCategoryTypeInput.STANDARD:
+                return
+        # If no capacity with type PARKING was found, we miss the capacity and therefore throw a validation error
+        raise ValidationError(reason='Missing parking spaces capacity')
+
+    def get_osm_opening_hours(self) -> str:
+        open_swiss_opening_times_by_weekday: dict[OpenDataSwissOperationTimeDaysOfWeek, list[str]] = defaultdict(list)
+        check_counter_24_7: int = 0
```
the-infinity, line 99: Why not just make a check for the `daysOfWeek` length?

src/parkapi_sources/converters/opendata_swiss/models.py
```diff
@@ -0,0 +1,200 @@
...
+from validataclass.exceptions import ValidationError
+from validataclass.validators import DataclassValidator, EnumValidator, IntegerValidator, ListValidator, Noneable, StringValidator
+
+from parkapi_sources.converters.base_converter.pull import GeojsonFeatureGeometryInput
+from parkapi_sources.models import RealtimeParkingSiteInput, StaticParkingSiteInput
+from parkapi_sources.models.enums import ParkAndRideType, ParkingSiteType, PurposeType
+
+
+@validataclass
+class OpenDataSwissAddressInput:
```
the-infinity, line 21: Are you sure about all the `Noneable`s in the following `validataclasses`? If I look at the data, most of it is well set and not `null`. For example, there is not a single `addressLine` which is `null`. Please have a look at the data and use `Noneable` just when it's needed.

src/parkapi_sources/converters/opendata_swiss/models.py
AbdullahiFatola: Now checked for none, but I am not sure if my approach is right by returning ```None``` for ```opening_hours```, if the inputs of open times are null.

src/parkapi_sources/converters/opendata_swiss/models.py
AbdullahiFatola: Now added ```GENERIC_BIKE````

src/parkapi_sources/converters/opendata_swiss/models.py
AbdullahiFatola: The whole method for ```get_osm_opening_hours``` have now been simplified

src/parkapi_sources/converters/opendata_swiss/models.py
AbdullahiFatola: Thanks for the hint

src/parkapi_sources/converters/opendata_swiss/converter.py
the-infinity: Without the permalink, you don't get any data. Just try to access the url now :)

It seems that Python requests automatically follows HTTP 302s, just checked it using a Python interpreter.

src/parkapi_sources/converters/opendata_swiss/models.py
```diff
@@ -0,0 +1,201 @@
...
+        opening_time: str = f'{self.operationTime.operatingFrom.strftime("%H:%M")}-{self.operationTime.operatingTo.strftime("%H:%M")}'
+        opening_time = '00:00-24:00' if opening_time == '00:00-00:00' else opening_time
+
+        for days_of_week_input in self.operationTime.daysOfWeek:
+            # If the weekday is within Monday to Friday, count to check later for OSM opening hours Mo-Fr
+            if days_of_week_input in list(OpenDataSwissOperationTimeDaysOfWeek)[:5]:
+                check_counter_weekday += 1
+
+        # If it's open all day and number of opening days is 7, then it is OSM - 24/7. No further handling needed in this case.
+        if opening_time == '00:00-24:00' and len(self.operationTime.daysOfWeek) == 7:
```
the-infinity, line 122: You can do this check earlier without collecting weekday: `if self.operationTime.operatingFrom == self.operationTime.operatingTo == time(0) and len(self.operationTime.daysOfWeek) == 7:`

src/parkapi_sources/converters/opendata_swiss/models.py
the-infinity: It's fine to set `opening_hours` to `None`, but you need to adapt the Typing.

Better way would be to use `__post_init__` in your validataclass: if just one of the two are set, throw a validation error.

src/parkapi_sources/converters/opendata_swiss/models.py
```diff
@@ -0,0 +1,201 @@
...
+        static_parking_site_input.opening_hours = self.properties.get_osm_opening_hours()
+
+        if self.properties.additionalInformationForCustomers:
+            static_parking_site_input.description = self.properties.additionalInformationForCustomers.de.replace('\n', ' ').replace(
+                '\r', ' '
+            )
+
+        if (
+            self.properties.address
+            and self.properties.address.addressLine
```
the-infinity, line 179: Why checking for the address.*? Do you want to catch emptystings, or ... ?

src/parkapi_sources/converters/opendata_swiss/models.py
```diff
@@ -0,0 +1,201 @@
...
+                check_counter_weekday += 1
+
+        # If it's open all day and number of opening days is 7, then it is OSM - 24/7. No further handling needed in this case.
+        if opening_time == '00:00-24:00' and len(self.operationTime.daysOfWeek) == 7:
+            return '24/7'
+
+        osm_opening_hour: list = []
+        # If the days are Monday to Friday with same opening time, we can summarize it to the Mo-Fr entry,
+        # otherwise we have to set it separately
+        if check_counter_weekday == 5:
```
the-infinity, line 128: Why not doing a direct check on the `daysOfWeek`? Like `all(day in self.operationTime.daysOfWeek for day in list(OpenDataSwissOperationTimeDaysOfWeek)[:5])` ? Then you don't need `check_counter_weekday` any more.

src/parkapi_sources/converters/opendata_swiss/models.py
```diff
@@ -0,0 +1,201 @@
...
+            lat=self.geometry.coordinates[1],
+            lon=self.geometry.coordinates[0],
+            static_data_updated_at=datetime.now(tz=timezone.utc),
+            has_realtime_data=False,
+        )
+
+        if self.properties.parkingFacilityCategory == PurposeType.CAR.value:
+            static_parking_site_input.purpose = PurposeType.CAR
+            static_parking_site_input.type = ParkingSiteType.CAR_PARK
+        elif self.properties.parkingFacilityCategory == PurposeType.BIKE.value:
```
the-infinity, line 163: I wonder why you check for BIKE at all?

src/parkapi_sources/converters/opendata_swiss/models.py
```diff
@@ -0,0 +1,201 @@
...
+@validataclass
+class OpenDataSwissPropertiesInput:
+    operator: str = StringValidator()
+    displayName: str = StringValidator()
+    address: Optional[OpenDataSwissAddressInput] = Noneable(DataclassValidator(OpenDataSwissAddressInput))
+    capacities: list[OpenDataSwissCapacitiesInput] = ListValidator(DataclassValidator(OpenDataSwissCapacitiesInput))
+    additionalInformationForCustomers: Optional[OpenDataSwissAdditionalInformationInput] = Noneable(
+        DataclassValidator(OpenDataSwissAdditionalInformationInput)
+    )
+    parkingFacilityCategory: str = StringValidator()
```
the-infinity, line 93: In my opinion, this can be an `EnumValidator`, or do you see any other value then `CAR`?

src/parkapi_sources/converters/opendata_swiss/models.py
AbdullahiFatola: Good tip, thanks

src/parkapi_sources/converters/opendata_swiss/models.py
AbdullahiFatola: There is an attribute ```bikeFacilityTraits``` that is always an empty list. In case of future when the portal includes "BIKE" in ```parkingFacilityCategory```. Plus we also pull the data from the portal, and the ```Änderungsdatum``` on the [portal ](https://opentransportdata.swiss/de/dataset/parking-facilities) changes, which I guess they update the data once in a while.

src/parkapi_sources/converters/opendata_swiss/models.py
```diff
@@ -0,0 +1,203 @@
...
+    operator: str = StringValidator()
+    displayName: str = StringValidator()
+    address: Optional[OpenDataSwissAddressInput] = Noneable(DataclassValidator(OpenDataSwissAddressInput))
+    capacities: list[OpenDataSwissCapacitiesInput] = ListValidator(DataclassValidator(OpenDataSwissCapacitiesInput))
+    additionalInformationForCustomers: Optional[OpenDataSwissAdditionalInformationInput] = Noneable(
+        DataclassValidator(OpenDataSwissAdditionalInformationInput)
+    )
+    parkingFacilityCategory: OpenDataSwissParkingFacilityCategory = EnumValidator(OpenDataSwissParkingFacilityCategory)
+    parkingFacilityType: Optional[str] = Noneable(StringValidator())
+    salesChannels: Optional[list[str]] = Noneable(ListValidator(StringValidator()))
```
the-infinity, line 121: I would either use this or remove it.

src/parkapi_sources/converters/opendata_swiss/models.py
```diff
@@ -0,0 +1,203 @@
...
+class OpenDataSwissPropertiesInput:
+    operator: str = StringValidator()
+    displayName: str = StringValidator()
+    address: Optional[OpenDataSwissAddressInput] = Noneable(DataclassValidator(OpenDataSwissAddressInput))
+    capacities: list[OpenDataSwissCapacitiesInput] = ListValidator(DataclassValidator(OpenDataSwissCapacitiesInput))
+    additionalInformationForCustomers: Optional[OpenDataSwissAdditionalInformationInput] = Noneable(
+        DataclassValidator(OpenDataSwissAdditionalInformationInput)
+    )
+    parkingFacilityCategory: OpenDataSwissParkingFacilityCategory = EnumValidator(OpenDataSwissParkingFacilityCategory)
+    parkingFacilityType: Optional[str] = Noneable(StringValidator())
```
the-infinity, line 120: Looks like an Enum, too

src/parkapi_sources/converters/opendata_swiss/models.py
```diff
@@ -0,0 +1,203 @@
...
+
+    def to_parking_site_type_input(self) -> ParkingSiteType:
+        # TODO: find out more details about this enumeration for a proper mapping
+        return {
+            self.CAR: ParkingSiteType.CAR_PARK,
+            self.BIKE: ParkingSiteType.GENERIC_BIKE,
+        }.get(self, ParkingSiteType.OTHER)
+
+    def to_purpose_type_input(self) -> ParkingSiteType:
+        # TODO: find out more details about this enumeration for a proper mapping
```
the-infinity, line 89: Is this really a TODO?

src/parkapi_sources/converters/opendata_swiss/models.py
```diff
@@ -0,0 +1,203 @@
...
+            self.SUNDAY: 'Su',
+        }.get(self, None)
+
+
+class OpenDataSwissParkingFacilityCategory(Enum):
+    CAR = 'CAR'
+    BIKE = 'BIKE'
+
+    def to_parking_site_type_input(self) -> ParkingSiteType:
+        # TODO: find out more details about this enumeration for a proper mapping
```
the-infinity, line 82: Is this really a TODO?

src/parkapi_sources/converters/opendata_swiss/models.py
```diff
@@ -0,0 +1,203 @@
...
+    def to_osm_opening_day_format(self) -> str:
+        return {
+            self.MONDAY: 'Mo',
+            self.TUESDAY: 'Tu',
+            self.WEDNESDAY: 'We',
+            self.THURSDAY: 'Th',
+            self.FRIDAY: 'Fr',
+            self.SATURDAY: 'Sa',
+            self.SUNDAY: 'Su',
+        }.get(self, None)
```
the-infinity, line 74: The `None` would mean that you return `None` if there is no key (and the typing would be wrong). As you described all keys, the `None` can just be removed.

src/parkapi_sources/converters/opendata_swiss/models.py
AbdullahiFatola: Now mapped to ```fee_description```

src/parkapi_sources/converters/opendata_swiss/models.py
AbdullahiFatola: Yeah that's right

PR #106, opened 2024-08-22
Integrated Hüfner Push Converter

AbdullahiFatola:
This PR adds the push converter for the static parking data from PARK SERVICE HÜFNER GmbH & Co. KG.


src/parkapi_sources/converters/huefner/converter.py
```diff
@@ -0,0 +1,54 @@
...
+            **goldbeck_header_rows,
+        }
+
+    def map_row_to_parking_site_dict(self, mapping: dict[str, int], row: list[Cell]) -> dict[str, Any]:
+        parking_site_dict = super().map_row_to_parking_site_dict(mapping, row)
+
+        for field in mapping.keys():
+            parking_site_dict[field] = row[mapping[field]].value
+
+        parking_site_dict['opening_hours'] = (
```
the-infinity, line 45: Why not doing a simple if without inline / else??

AbdullahiFatola: Oh that's right!

src/parkapi_sources/converters/huefner/converter.py
```diff
@@ -0,0 +1,52 @@
...
+            **goldbeck_header_rows,
+        }
+
+    def map_row_to_parking_site_dict(self, mapping: dict[str, int], row: list[Cell]) -> dict[str, Any]:
+        parking_site_dict = super().map_row_to_parking_site_dict(mapping, row)
+
+        for field in mapping.keys():
+            parking_site_dict[field] = row[mapping[field]].value
+
+        if '-00:00' in parking_site_dict['opening_hours']:
```
the-infinity, line 45: Even simpler: `endswith` :)

AbdullahiFatola: Yeah, i also thought about ```endswith``` but in the case of opening times like```Mo-Fr 06:00-00:00; Sa 08:00-23:00```, then the IF condition will be ignored and ```00:00``` will not be replaced with ```24:00```, since the string does not end with ```00:00```. Or what do you think?

the-infinity: Ah! Ok, you are right, did not think about multiple day. Then it's fine your way.

PR #107, opened 2024-08-22
APCOA Converter: Fixes the OSM opening_hours (24:00)

AbdullahiFatola:
This PR fixes the APCOA ```opening_hours``` ending with 00:00 by replacing it with OSM Format 24:00 (e.g. Parking place with UID: ```20236``` with opening times - ```Mo 05:00-00:00; Tu 05:00-00:00; We 05:00-00:00; Th 05:00-02:00; Fr 05:00-02:00; Sa 05:00-02:00; Su 05:00-02:00)``` and adding a test script to ensure all opening times conforms with OSM Format. 

In addition the ```photo_url``` field and values were removed from the dataset, because the parking place photos ```CarParkPhotoURLs``` of APCOA data contain pictures in Blob Format ([Example parking](https://apcoatx.blob.core.windows.net/mdm/fc44f33f-d70f-44f1-8418-ab1640fc39dc)). The photos does not load directly with the URL. To view the photo, it must be downloaded and opened with a photo app or photo software, which does not conform with purpose of the ```photo_url``` data model in the ParkAPI.

### Updated:
```photo_url``` was added back since the values of ```CarParkPhotoURLs``` in APCOA Data are valid JPEG and can work perfectly for data users in an application or in HTML tag.

### DIFF
--- a/src/parkapi_sources/converters/apcoa/mapper.py
+++ b/src/parkapi_sources/converters/apcoa/mapper.py
@@ -44,9 +44,6 @@ class ApcoaMapper:
         if apcoa_input.Address.Street and apcoa_input.Address.Zip and apcoa_input.Address.City:
             static_parking_site_input.address = f'{apcoa_input.Address.Street}, {apcoa_input.Address.Zip} {apcoa_input.Address.City}'
 
-        if apcoa_input.CarParkPhotoURLs:
-            static_parking_site_input.photo_url = apcoa_input.CarParkPhotoURLs.CarparkPhotoURL1
-
         if apcoa_input.IndicativeTariff.MinValue or apcoa_input.IndicativeTariff.MaxValue:
             static_parking_site_input.has_fee = True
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by the-infinity** [comment]
The `photo_url` is actually a valid JPEG which works perfectly fine in an HTML:

```
<html><body><img src="https://apcoatx.blob.core.windows.net/mdm/fc44f33f-d70f-44f1-8418-ab1640fc39dc"></body></html>
``` 

So these photos are fine for `photo_url`. If you access the photo directly, the browser does not know what this file is, because the server does a generic `Content-Type: application/octet-stream` instead of `image/jpeg`, but if you add this context using HTML, it works fine.

AbdullahiFatola:
Okay does this mean the data users have to embed the URL in an HTML tag for usage on their side?

the-infinity:
Yea, or anything else which says that this is an image. There's no redirect or anything, just a wrong = too generic mimetype, so the browser does not know what this file is if nobody gives context. Will work perfectly in any application where there is an HTML img tag.

AbdullahiFatola:
Ok thanks for clarifying.

Issue #139, opened 2024-09-27
pbw attributes


ThorstenFroehlinghaus:
pwd parking sites with art_lang="Parkbereich" shall be mapped to OFF_STREET_PARKING_GROUND instead of ON_STREET.
pwd parking sites shall obtain an attribute public_url having the form "https://www.pbw.de/?menu=parkplatz-finder&search=*uid"

the-infinity:
Implemented by https://github.com/ParkenDD/parkapi-sources-v3/pull/144

PR #144, opened 2024-10-08
#139 - Update data for static parking sites and expand test to check …

JGaukrogers:
…converted data from one site


the-infinity (approved):

the-infinity:
... sadly, ruff fails here. You can use the pre-commit-hook to check the code before you commit, then you don't run into these issues.

Issue #161, opened 2024-10-30
Use static data as additional data input for kienzler


the-infinity:
With https://github.com/ParkenDD/parkapi-static-data/pull/6 , there will be a geojson which should be used as mixin for Kienzler boxes at the static data handling. Please extend the static Kienzler base converter with the ability to get these GeoJSONs, and mix the properties into the properties from Kienzlers API.

If the GeoJSON should be fetched and mixed in at all should be controlled by a property of the base class, which should be false by default, and can be activated in the subclasses.

Please re-use the existing GeoJSON validators when implementing this feature.

Please use `https://raw.githubusercontent.com/ParkenDD/parkapi-static-data/refs/heads/main/sources/{source.uid}.json` as GeoJSON URL.

If a dataset exists in Kienzler API, but not in the GeoJSON, that's fine, please add the dataset anyway, just without the extended data from the GeoJSON. If it's the other way around, dataset in GeoJSON, but not in Kienzler, please ignore this dataset.

the-infinity:
Implemented with version 0.15.0

PR #168, opened 2024-11-14
161 use static data as additional data input for kienzler

JGaukrogers:
Solves issue #161: https://github.com/ParkenDD/parkapi-sources-v3/issues/161

### DIFF
--- a/src/parkapi_sources/converters/kienzler/converter.py
+++ b/src/parkapi_sources/converters/kienzler/converter.py
@@ -20,21 +20,18 @@ class KienzlerBasePullConverter(PullConverter, StaticGeojsonDataMixin):
     kienzler_list_validator = ListValidator(AnythingValidator(allowed_types=[dict]))
     kienzler_item_validator = DataclassValidator(KienzlerInput)
     use_geojson = False
+    geojson_feature_validator = DataclassValidator(KienzlerGeojsonFeatureInput)
 
     @property
     @abstractmethod
     def config_prefix(self):
         pass
 
-    def __init__(self, *args, use_geojson=False, **kwargs):
-        super().__init__(*args, **kwargs)
-        self.required_config_keys = [
-            f'PARK_API_KIENZLER_{self.config_prefix}_USER',
-            f'PARK_API_KIENZLER_{self.config_prefix}_PASSWORD',
-            f'PARK_API_KIENZLER_{self.config_prefix}_IDS',
-        ]
-        self.use_geojson = use_geojson
-        self.geojson_feature_validator = DataclassValidator(KienzlerGeojsonFeatureInput)
+    required_config_keys = [
+        f'PARK_API_KIENZLER_{config_prefix}_USER',
+        f'PARK_API_KIENZLER_{config_prefix}_PASSWORD',
+        f'PARK_API_KIENZLER_{config_prefix}_IDS',
+    ]
 
     def get_static_parking_sites(self) -> tuple[list[StaticParkingSiteInput], list[ImportParkingSiteException]]:
         kienzler_parking_sites, static_parking_site_errors = self._get_kienzler_parking_sites()
@@ -159,6 +156,7 @@ class KienzlerNeckarsulmPullConverter(KienzlerBasePullConverter):
 
 class KienzlerOffenburgPullConverter(KienzlerBasePullConverter):
     config_prefix = 'OFFENBURG'
+    use_geojson = True
 
     source_info = SourceInfo(
         uid='kienzler_offenburg',
@@ -171,6 +169,7 @@ class KienzlerOffenburgPullConverter(KienzlerBasePullConverter):
 
 class KienzlerRadSafePullConverter(KienzlerBasePullConverter):
     config_prefix = 'RADSAFE'
+    use_geojson = True
 
     source_info = SourceInfo(
         uid='kienzler_rad_safe',
@@ -183,6 +182,7 @@ class KienzlerRadSafePullConverter(KienzlerBasePullConverter):
 
 class KienzlerStuttgartPullConverter(KienzlerBasePullConverter):
     config_prefix = 'STUTTGART'
+    use_geojson = True
 
     source_info = SourceInfo(
         uid='kienzler_stuttgart',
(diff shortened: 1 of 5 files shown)


src/parkapi_sources/converters/kienzler/converter.py
```diff
@@ -9,36 +9,59 @@
...
+    kienzler_geojson_item_validator = DataclassValidator(KienzlerInput)
+    use_geojson = False
 
     @property
     @abstractmethod
     def config_prefix(self):
         pass
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, use_geojson=False, **kwargs):
```
the-infinity, line 30: Missing typing. But: you cannot set this attribute anyway. Instead, it should be set at the class definition as you already do above.

Now it's also clear which subclasses should set `se_geojson = True`. It's Offenburg, RadSafe, Stuttgart and VRN (see https://github.com/ParkenDD/parkapi-static-data/pull/6/files ).

src/parkapi_sources/converters/kienzler/converter.py
```diff
@@ -9,36 +9,59 @@
...
+        if self.use_geojson:
+            geojson_parking_sites, geojson_parking_site_errors = self._get_static_geojson_parking_sites()
+
+            static_geojson_parking_site_inputs_by_uid: dict[str, KienzlerGeojsonFeatureInput] = {}
+            for geojson_parking_site in geojson_parking_sites:
+                static_geojson_parking_site_inputs_by_uid[geojson_parking_site.uid] = geojson_parking_site
+
+            static_parking_site_errors += geojson_parking_site_errors
+
+            # For each GeoJSON entry, extend its correspondent static data
```
the-infinity, line 56: I think this comment is outdated: it's for each Kienzler entry, extend it with GeoJSON feature if the feature exists

src/parkapi_sources/converters/kienzler/converter.py
```diff
@@ -9,36 +9,59 @@
...
 
-from .models import KienzlerInput
+from .models import KienzlerGeojsonFeatureInput, KienzlerInput
 
 
-class KienzlerBasePullConverter(PullConverter):
+class KienzlerBasePullConverter(PullConverter, StaticGeojsonDataMixin):
     kienzler_list_validator = ListValidator(AnythingValidator(allowed_types=[dict]))
     kienzler_item_validator = DataclassValidator(KienzlerInput)
+    kienzler_geojson_item_validator = DataclassValidator(KienzlerInput)
```
the-infinity, line 22: I don't see any place where you use this

src/parkapi_sources/converters/kienzler/converter.py
```diff
@@ -9,36 +9,59 @@
...
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, use_geojson=False, **kwargs):
         super().__init__(*args, **kwargs)
         self.required_config_keys = [
             f'PARK_API_KIENZLER_{self.config_prefix}_USER',
             f'PARK_API_KIENZLER_{self.config_prefix}_PASSWORD',
             f'PARK_API_KIENZLER_{self.config_prefix}_IDS',
         ]
+        self.use_geojson = use_geojson
+        self.geojson_feature_validator = DataclassValidator(KienzlerGeojsonFeatureInput)
```
the-infinity, line 38: Why not initializing this outside of __init__? (eg where you most likely will delete `kienzler_geojson_item_validator`)?

src/parkapi_sources/converters/kienzler/models.py
```diff
@@ -45,3 +57,60 @@ def to_realtime_parking_site(self) -> RealtimeParkingSiteInput:
...
+    type: Optional[ParkingSiteType] = EnumValidator(ParkingSiteType), Default(None)
+    max_height: int = IntegerValidator(min_value=0)
+    max_width: int = IntegerValidator(min_value=0)
+    max_depth: int = IntegerValidator(min_value=0)
+    park_and_ride_type: list[ParkAndRideType] = ListValidator(EnumValidator(ParkAndRideType))
+    external_identifiers: Optional[list[ExternalIdentifier]] = ListValidator(DataclassValidator(ExternalIdentifier))
+
+
+@validataclass
+class KienzlerStaticParkingSiteInput(BaseParkingSiteInput):
```
the-infinity, line 81: I do wonder why we need that class at all. In the end, you put data in the class at `to_static_parking_site_input()`, and make a dict again at `kienzler_parking_site.from_dict(static_geojson_parking_site_inputs_by_uid[parking_site_uid].to_dict())`. Why not simplify it by renaming `to_static_parking_site_input' to `to_static_parking_site_input_dicts`, return a dict there like this: `return {'lat': self.geometry.coordinates[1],'lon': self.geometry.coordinates[0],**properties_dict}`, `static_geojson_parking_site_inputs_by_uid` gets a `dict[str, dict]` then, which means you can load the dict directly at `from_dict`: `kienzler_parking_site.from_dict(static_geojson_parking_site_inputs_by_uid[parking_site_uid])`? Or do I miss something why `KienzlerStaticParkingSiteInput` is needed?

JGaukrogers: How did you notice this? :O

src/parkapi_sources/converters/kienzler/converter.py
JGaukrogers: Done

src/parkapi_sources/converters/kienzler/converter.py
```diff
@@ -9,36 +9,55 @@
...
         pass
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self.required_config_keys = [
-            f'PARK_API_KIENZLER_{self.config_prefix}_USER',
-            f'PARK_API_KIENZLER_{self.config_prefix}_PASSWORD',
-            f'PARK_API_KIENZLER_{self.config_prefix}_IDS',
-        ]
+    required_config_keys = [
```
the-infinity, line 30: Sadly, this does not work, because `config_prefix` is just a abstract property in this moment. So `required_config_keys` is `['PARK_API_KIENZLER_<property object at 0x7935a5ba0f40>_USER', 'PARK_API_KIENZLER_<property object at 0x7935a5ba0f40>_PASSWORD', 'PARK_API_KIENZLER_<property object at 0x7935a5ba0f40>_IDS']` if you do this. The old approach setting these values at init was the correct one, because there it was set during (child) object initialization, where `config_prefix` was already set.

PR #169, opened 2024-11-18
Radvis_bw: Add Enum BILDUNGSEINRICHTUNG to related_location 

AbdullahiFatola:
This PR adds the new enum from the RadVIS data model - ```BILDUNGSEINRICHTUNG``` to the field ```related_location``` in the ```radvis_bw``` source converter.


src/parkapi_sources/converters/radvis_bw/converter.py
```diff
@@ -38,7 +38,7 @@ def get_static_parking_sites(self) -> tuple[list[StaticParkingSiteInput], list[I
         static_parking_site_errors: list[ImportParkingSiteException] = []
 
         parking_site_features = self.geojson_validator.validate(self.get_data())
-        sources_to_ignore: list[str] = []
+        sources_to_ignore: list[str] = ['MOBIDATABW']
```
the-infinity, line 41: That should be part of the config as loaded below, not hardcoded in this converter.

AbdullahiFatola: Now set in the unit test.

the-infinity (approved):

PR #172, opened 2024-11-25
Add bahnv2 bike parking locked and open

AbdullahiFatola:
This PR updates existing ```bahnv2``` parking source converter by integrating the new capacity types ```BIKE_PARKING_LOCKED``` and ```BIKE_PARKING_OPEN``` as separate ```ParkingSite``` with ```type``` of ```LOCKBOX``` and ```OTHER``` respectively for bike parking. The existing mappings for car parking ```type``` are not changed. 

The parking sites from Deutsche BahnPark ```bahnv2``` now have ```purpose``` as ```bike``` and ```car``` , grouped into ```ParkingSiteGroup```.


src/parkapi_sources/converters/bahn_v2/mapper.py
```diff
@@ -26,6 +33,13 @@ def map_static_parking_site(bahn_input: BahnParkingSiteInput) -> StaticParkingSi
             has_realtime_data=False,  # TODO: change this as soon as Bahn offers proper rate limits
             static_data_updated_at=datetime.now(tz=timezone.utc),
             public_url=bahn_input.url,
+            purpose=next(
```
the-infinity, line 36: Why looping here at all? I mean, you select the type `BahnParkingSiteCapacityType.PARKING` anyway?

src/parkapi_sources/converters/bahn_v2/mapper.py
```diff
@@ -44,3 +58,93 @@ def map_static_parking_site(bahn_input: BahnParkingSiteInput) -> StaticParkingSi
                 static_parking_site_input.capacity_disabled = int(round(capacity_data.total))
 
         return static_parking_site_input
+
+
+class BahnBikeParkingLockedMapper:
```
the-infinity, line 63: That and the mapper below is almost a 100 % copy of the mapper above. In general, please try to avoid copy-pasting large parts of the code and do OOP. In this case, I wonder why you need different Mappers at all, as this could all be done in one source and one mapper.

src/parkapi_sources/converters/bahn_v2/validators.py
```diff
@@ -137,3 +173,43 @@ def __post_init__(self):
                 return
         # If no capacity with type PARKING was found, we miss the capacity and therefore throw a validation error
         raise ValidationError(reason='Missing parking capacity')
+
+
+@validataclass
+class BahnBikeLockedParkingSiteInput:
```
the-infinity, line 179: Same as Mappers: copy paste, and can be removed completely as a single source would work, too.

src/parkapi_sources/converters/bahn_v2/converter.py
```diff
@@ -70,3 +60,39 @@ def get_data(self) -> dict:
             timeout=60,
         )
         return response.json()
+
+
+class BahnV2PullConverter(BahnV2BasePullConverter):
```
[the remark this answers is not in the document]
AbdullahiFatola: You are right with the rate limit!

src/parkapi_sources/converters/bahn_v2/mapper.py
AbdullahiFatola: It's now updated.

src/parkapi_sources/converters/bahn_v2/validators.py
AbdullahiFatola: Updated, thanks!

src/parkapi_sources/converters/bahn_v2/mapper.py
```diff
@@ -6,41 +6,69 @@
...
+
+    @staticmethod
+    def map_static_parking_site_bike_open(bahn_input: BahnParkingSiteInput) -> StaticParkingSiteInput:
+        return BahnMapper._map_static_parking_site(
+            bahn_input, '-bike-open', PurposeType.BIKE, BahnParkingSiteCapacityType.BIKE_PARKING_OPEN
+        )
+
+    @staticmethod
+    def _map_static_parking_site(
+        bahn_input, uid_suffix: str, purpose: PurposeType, capacity_type: BahnParkingSiteCapacityType
```
the-infinity, line 39: `bahn_input` needs typing (and the whole line a trailing comma)

src/parkapi_sources/converters/bahn_v2/mapper.py
```diff
@@ -6,41 +6,69 @@
...
                 iter(name_input.name for name_input in bahn_input.name if name_input.context == NameContext.NAME)
             ),
             lat=bahn_input.address.location.latitude,
             lon=bahn_input.address.location.longitude,
             operator_name=bahn_input.operator.name,
             address=f'{bahn_input.address.streetAndNumber}, {bahn_input.address.zip} {bahn_input.address.city}',
-            type=bahn_input.type.name.to_parking_site_type_input(),
+            type=next(
+                iter(
+                    item.to_bike_parking_site_type_input() for item in bahn_input.capacity if item.type == capacity_type
```
the-infinity, line 53: The `next(iter(...`  lines are a bit difficult to understand, and you do the same `next(iter(` iteration multiple times. I would suggest to add a method `BahnParkingSiteInput.get_capacity_by_type(capacity_type: BahnParkingSiteCapacityType)`, which returns a `BahnCapacityInput`, and then using this method before `static_parking_site_input = StaticParkingSiteInput(` just one time. This line would look like `type=capacity_input.to_bike_parking_site_type_input()` afterwards.

src/parkapi_sources/converters/bahn_v2/mapper.py
```diff
@@ -6,41 +6,69 @@
...
+                )
+            )
+            if capacity_type != BahnParkingSiteCapacityType.PARKING
```
the-infinity, line 56: You can put this below, which would remove the inline if which is not in one line any more because of the length.

src/parkapi_sources/converters/bahn_v2/validators.py
```diff
@@ -127,13 +162,13 @@ class BahnParkingSiteInput:
     type: BahnTypeInput = DataclassValidator(BahnTypeInput)
     operator: BahnOperatorInput = DataclassValidator(BahnOperatorInput)
     address: BahnAdressInput = DataclassValidator(BahnAdressInput)
+    station: BahnStationInput = DataclassValidator(BahnStationInput)
```
the-infinity, line 165: Why did you add this validator? I mean, it's fine if you want to use it afterwards, but then you need to add the data at your `BahnMapper`

src/parkapi_sources/converters/bahn_v2/mapper.py
AbdullahiFatola: thanks for the tip!

src/parkapi_sources/converters/bahn_v2/validators.py
AbdullahiFatola: oops...I thought I will be using it for group_uid before, but not anymore. I have now removed it.

the-infinity:
Added some changes to make sure that there is always a valid value and rebased it.

PR #173, opened 2024-11-27
more abstract geojson approach

the-infinity:
As https://github.com/ParkenDD/parkapi-sources-v3/pull/168 was running into strange errors during testing, I made the GeoJSON approach a bit more abstract and moved it to a general model.


src/parkapi_sources/converters/herrenberg_bike/models.py
```diff
@@ -112,7 +111,8 @@ class HerrenbergBikePropertiesInput(ValidataclassMixin):
     operator_name: OptionalUnset[str] = NoneToUnsetValue(StringValidator(min_length=0, max_length=256)), DefaultUnset
     capacity: int = IntegerValidator(min_value=0)
     capacity_charging: OptionalUnset[int] = NoneToUnsetValue(IntegerValidator(min_value=0)), DefaultUnset
-    capacity_cargobike: OptionalUnset[int] = NoneToUnsetValue(IntegerValidator(min_value=0)), DefaultUnset
+    # capacity_cargobike is unsupported yet
```
JGaukrogers, line 114: Question: is it known when it will be supported? I am wondering if it would help to write the comment as a TODO + number of the issue where it will be fixed, so it's easier to find it later.

src/parkapi_sources/converters/herrenberg_bike/models.py
```diff
@@ -145,36 +146,32 @@ class HerrenbergBikePropertiesInput(ValidataclassMixin):
     max_stay: OptionalUnset[int] = NoneToUnsetValue(IntegerValidator(min_value=0)), DefaultUnset
     fee_description: OptionalUnset[str] = NoneToUnsetValue(StringValidator(max_length=512)), DefaultUnset
 
+    def to_dict(self, *args, **kwargs) -> dict:
+        ignore_keys = ['type', 'original_uid', 'supervision_type', 'date_surveyed']
+        result = {key: value for key, value in super().to_dict(*args, **kwargs).items() if key not in ignore_keys}
```
JGaukrogers, line 151: Minor issue: I am thinking about if renaming `result` as `new_dictionary` or something like that would make the code a bit better understandable.

src/parkapi_sources/converters/kienzler/converter.py
```diff
@@ -90,7 +90,7 @@ def _get_kienzler_parking_sites(self) -> tuple[list[KienzlerInput], list[ImportP
                 )
         return kienzler_item_inputs, errors
 
-    def _request(self) -> list[dict]:
+    def _request_kienzler(self) -> list[dict]:
```
JGaukrogers, line 93: Just out of curiosity: was the renaming done because `_request_kienzler` is a better name, or is there another reason, like `_request` already existing for something else?

src/parkapi_sources/converters/kienzler/models.py
```diff
@@ -71,19 +65,21 @@ class KienzlerGeojsonFeaturePropertiesInput(ValidataclassMixin):
     max_width: int = IntegerValidator(min_value=0)
     max_depth: int = IntegerValidator(min_value=0)
     park_and_ride_type: list[ParkAndRideType] = ListValidator(EnumValidator(ParkAndRideType))
-    external_identifiers: Optional[list[ExternalIdentifier]] = ListValidator(DataclassValidator(ExternalIdentifier))
+    external_identifiers: Optional[list[ExternalIdentifierInput]] = ListValidator(
+        DataclassValidator(
+            ExternalIdentifierInput,
+        ),
```
JGaukrogers, line 71: Suggestion: write `DataclassValidator(ExternalIdentifierInput),` instead.

src/parkapi_sources/models/geojson_inputs.py
```diff
@@ -21,12 +21,23 @@
...
-from parkapi_sources.models.enums import ParkingSiteType
+from .enums import ParkingSiteType
+from .parking_site_inputs import StaticParkingSiteInput
 
 
 @validataclass
-class GeojsonFeaturePropertiesInput(ValidataclassMixin):
+class GeojsonBaseFeaturePropertiesInput(ValidataclassMixin):
+    def to_dict(self, *args, static_data_updated_at: datetime | None = None, **kwargs) -> dict:
+        result = super().to_dict()
```
JGaukrogers, line 31: Suggestion: rename `result` to `feature_dictionary` or something like that, because that is more precise on what this variable is supposed to be. However the method is so small, that it's also ok for me if you leave it as it is.

src/parkapi_sources/converters/herrenberg_bike/models.py
the-infinity: Did not do a TODO, as it is no TODO atm, but linked the ticket: https://github.com/ParkenDD/parkapi-sources-v3/issues/174

src/parkapi_sources/converters/herrenberg_bike/models.py
the-infinity: I would keep it that way as a shorter variable name helps to stay in one line which keeps overview, and it is the result of this methos, so the name is not wrong

src/parkapi_sources/converters/kienzler/converter.py
the-infinity: Because we have two requests there, one to Kinezler and one to the GeoJSON. And confused myself multiple times because I had to look in the method which data is requested.

src/parkapi_sources/models/geojson_inputs.py
the-infinity: Same as above: it is the result, and result is shorter.

PR #191, opened 2025-02-25
Fix and reduce validation errors in converters

AbdullahiFatola:
This PR fixes validation errors in some converters. These fixes only apply to the data models and attribute mappings in the converters, and are only taken into consideration, If the data are plausible from the sources. In situation where the data is not plausible, then we inform the contact person of the Data provider to fix the data from the source.


src/parkapi_sources/converters/base_converter/push/normalized_xlsx_converter.py
```diff
@@ -111,8 +111,17 @@ def map_row_to_parking_site_dict(
         opening_hours_input = self.excel_opening_time_validator.validate({
             key: value for key, value in parking_site_raw_dict.items() if key.startswith('opening_hours_')
         })
+
+        for boolean_field in ['has_fee', 'is_covered', 'has_lighting']:
```
the-infinity, line 115: This should be part of the validator. There are some examples for validator children in the project, but if child-classing validators is too overwhelming, I can do that, too.

src/parkapi_sources/converters/heidelberg/models.py
```diff
@@ -86,40 +86,62 @@ class HeidelbergParkingSubType(Enum):
 
 @validataclass
 class HeidelbergInput:
-    acceptedPaymentMethod: HeidelbergPaymentMethodType = RemoveValueDict(
-        ListValidator(EnumValidator(HeidelbergPaymentMethodType)),
+    id: str = StringValidator()
+    acceptedPaymentMethod: Optional[HeidelbergPaymentMethodType] = (
```
the-infinity, line 90: Typing is wrong here, it's `list[HeidelbergPaymentMethodType]`.

src/parkapi_sources/converters/heidelberg/models.py
```diff
@@ -86,40 +86,62 @@ class HeidelbergParkingSubType(Enum):
...
-    description: str = RemoveValueDict(ReplacingStringValidator(mapping={'\r': '', '\n': ' ', '\xa0': ' '}))
-    facilities: list[str] = RemoveValueDict(ListValidator(EnumValidator(HeidelbergFacilityType)))
-    familyParkingSpots: int = RemoveValueDict(IntegerValidator())
-    googlePlaceId: str = RemoveValueDict(StringValidator())
-    handicappedParkingSpots: int = RemoveValueDict(IntegerValidator())
-    images: list[str] = RemoveValueDict(ListValidator(UrlValidator()))
+    addressLocality: Optional[str] = NoneableRemoveValueDict(StringValidator()), Default(None)
+    availableSpotNumber: Optional[int] = NoneableRemoveValueDict(IntegerValidator(min_value=0)), Default(None)
+    closingHours: Optional[time] = (
+        NoneableRemoveValueDict(TimeValidator(time_format=TimeFormat.NO_SECONDS)),
```
the-infinity, line 99: Any cases where you actually need the `NoneableRemoveValueDict`, and not `Noneable`, or even nothing of it, because if it's set, it's never None? Don't see them in test data, but maybe I miss something?

src/parkapi_sources/converters/radvis_bw/models.py
```diff
@@ -82,6 +82,7 @@ def to_parking_site_type(self) -> ParkingSiteType:
 
 class StatusType(Enum):
     AKTIV = 'AKTIV'
+    GEPLANT = 'GEPLANT'
```
the-infinity, line 85: Is that by purpose that we still store these parking sites without any remark? Maybe they should be ignored in the converter?

src/parkapi_sources/converters/vrn_p_r/models.py
```diff
@@ -174,7 +174,9 @@ def to_realtime_parking_site_input(self) -> RealtimeParkingSiteInput:
 
         return RealtimeParkingSiteInput(
             uid=f'{self.properties.original_uid}-{self.properties.vrn_sensor_id}',
-            realtime_capacity=self.properties.realtime_free_capacity + self.properties.realtime_occupied,
+            realtime_capacity=self.properties.realtime_free_capacity + self.properties.realtime_occupied
```
the-infinity, line 177: Please don't make this inline, it's pretty hard to read.

Also, you can do `self.properties.realtime_free_capacity is not UnsetValue`, as `UnsetValue` is a [sentinel object](https://python-patterns.guide/python/sentinel-object/), that's more efficient.

the-infinity (commented):
Good work to handle all these errors!

src/parkapi_sources/converters/base_converter/push/normalized_xlsx_converter.py
AbdullahiFatola: Now added to the validator, thanks for the hint!

src/parkapi_sources/converters/radvis_bw/models.py
AbdullahiFatola: It is now made ```Optional``` to ignore in converter and avoid validation error.

src/parkapi_sources/converters/vrn_p_r/models.py
AbdullahiFatola: Good tip!

src/parkapi_sources/converters/radvis_bw/models.py
AbdullahiFatola: okay I would just validate them and ignore in final ParkingSites, since there are existing parking lots having ```StatusType.GEPLANT```` and that could mislead the clients driving to the location, if they get to the location and the parking lot is under construction or not usable.

the-infinity (approved):

PR #198, opened 2025-03-26
friedrichhafen sensors, parking spot model

the-infinity:
Adds friedrichhafen sensors source, which is the first parking spot source, therefore we new validations and models included


docs/mapping/friedrichshafen_sensors.md
```diff
@@ -9,6 +9,8 @@ Attributes which are set statically:
 
 * `realtime_opening_status` is always set to `OPEN`
```
AbdullahiFatola, line 10: `realtime_opening_status` seems not be in `RealtimeParkingSpotInput`, i found only `realtime_status` or did I missing something?

docs/mapping/friedrichshafen_sensors.md
```diff
@@ -9,6 +9,8 @@ Attributes which are set statically:
...
 * `has_realtime_data` is always set to `true`
+* `purpose` is always set to `CAR`
+* `has_realtime_data` is always set to `true`
```
AbdullahiFatola, line 13: `has_realtime_data` seems to be duplicated.

src/parkapi_sources/models/parking_spot_inputs.py
```diff
@@ -0,0 +1,81 @@
...
+        ),
+        min_length=1,
+        max_length=1,
+    )
+
+
+@validataclass
+class StaticParkingSpotInput(ValidataclassMixin):
+    uid: str = StringValidator(min_length=1, max_length=256)
+    name: str = StringValidator(min_length=1, max_length=256)
```
AbdullahiFatola, line 47: If I am right, `name` is not required in the new data model for `ParkingSpot`, but it is defined here as required. Would this not lead to `ValidationError` in other converters If the `ParkingSpot` has no name?

src/parkapi_sources/converters/friedrichshafen_sensors/validators.py
```diff
@@ -0,0 +1,57 @@
...
+from parkapi_sources.models.parking_restriction_inputs import ParkingRestrictionInput
+
+
+@validataclass
+class FriedrichshafenSensorsParkingSpot(UrbanParkingSite):
+    # As Friedrichshafen uses a parking site model for parking spots, we have to ensure that the capacity is always 1
+    parkingNumberOfSpaces: int = AnyOfValidator(allowed_values=['1'])
+
+    def to_static_parking_spot_input(self) -> StaticParkingSpotInput:
+        name_de = ''
```
AbdullahiFatola, line 23: Since `name` is not required in the `ParkingSpot` model, Would it not be better If is optional or ignored instead of having it as empty string?

AbdullahiFatola (commented):
Thank you for this template. It will really help in the developing the Ulm converter. I just have few observations and clarifications in the review.

docs/mapping/friedrichshafen_sensors.md
the-infinity: Oh, right, that was ParkingSite. Removed it.

docs/mapping/friedrichshafen_sensors.md
the-infinity: Removed it.

src/parkapi_sources/models/parking_spot_inputs.py
the-infinity: You are right. Fixed.

src/parkapi_sources/converters/friedrichshafen_sensors/validators.py
the-infinity: Good idea.

src/parkapi_sources/converters/friedrichshafen_sensors/converter.py
```diff
@@ -32,7 +38,7 @@ def get_static_parking_spots(self) -> tuple[list[StaticParkingSpotInput], list[I
         )
 
         static_parking_spot_inputs: list[StaticParkingSpotInput] = []
-        static_parking_site_errors: list[ImportParkingSiteException] = []
+        static_parking_spot_errors: list[ImportParkingSiteException] = []
```
AbdullahiFatola, line 42: I discovered there is `ImportParkingSpotException`, i think we should also use it here.

src/parkapi_sources/converters/friedrichshafen_sensors/converter.py
```diff
@@ -44,15 +50,16 @@ def get_static_parking_spots(self) -> tuple[list[StaticParkingSpotInput], list[I
                 static_parking_spot_inputs.append(static_parking_spot_input)
 
             except ValidationError as e:
-                static_parking_site_errors.append(
+                static_parking_spot_errors.append(
                     ImportParkingSiteException(
                         source_uid=self.source_info.uid,
                         parking_site_uid=self.get_uid_from_static_input_dict(static_input_dict),
```
AbdullahiFatola, line 56: If we use `ImportParkingSpotException` then `parking_spot_uid` would be suitable here.

src/parkapi_sources/converters/friedrichshafen_sensors/converter.py
```diff
@@ -6,6 +6,7 @@
 from validataclass.exceptions import ValidationError
 from validataclass.validators import DataclassValidator
 
+from parkapi_sources.converters.base_converter import ParkingSpotBaseConverter
 from parkapi_sources.converters.base_converter.datex2 import ParkingRecordStatusMixin, UrbanParkingSiteMixin
 from parkapi_sources.converters.base_converter.pull import MobilithekPullConverter
 from parkapi_sources.exceptions import ImportParkingSiteException
```
AbdullahiFatola, line 12: Would `ImportParkingSpotException` not be better here? or what do you think?

src/parkapi_sources/converters/friedrichshafen_sensors/converter.py
```diff
@@ -44,15 +50,16 @@ def get_static_parking_spots(self) -> tuple[list[StaticParkingSpotInput], list[I
                 static_parking_spot_inputs.append(static_parking_spot_input)
 
             except ValidationError as e:
-                static_parking_site_errors.append(
+                static_parking_spot_errors.append(
                     ImportParkingSiteException(
```
AbdullahiFatola, line 54: Here also `ImportParkingSpotException`?

src/parkapi_sources/converters/friedrichshafen_sensors/converter.py
```diff
@@ -44,15 +50,16 @@ def get_static_parking_spots(self) -> tuple[list[StaticParkingSpotInput], list[I
...
                         parking_site_uid=self.get_uid_from_static_input_dict(static_input_dict),
                         message=str(e.to_dict()),
+                        data=static_input_dict,
                     ),
                 )
 
-        return static_parking_spot_inputs, static_parking_site_errors
+        return static_parking_spot_inputs, static_parking_spot_errors
 
     def get_realtime_parking_spots(self) -> tuple[list[RealtimeParkingSpotInput], list[ImportParkingSiteException]]:
```
AbdullahiFatola, line 64: Also here `ImportParkingSpotException`?

AbdullahiFatola, line 64: Also here ImportParkingSpotException?

tests/converters/friedrichshafen_sensors_test.py
AbdullahiFatola, line 1: Ì just noticed the functions and parameters here contain `...parking_site...`. I would use `...parking_spot...` instead

src/parkapi_sources/converters/friedrichshafen_sensors/converter.py
AbdullahiFatola, line 1: Since you already defined `ImportParkingSpotException` in `exceptions.py`, I think it would make sense to use in this converter instead of `ImportParkingSiteException` or what do you think? 

Likewise `parking_spot_uid` in place of `parking_site_uid` and `realtime_parking_spot_errors` in place of `realtime_parking_site_errors`.

AbdullahiFatola (commented):
I think we need to rename some functions, variables and classes in the `friedrichshafen_sensors` converter and its test converter.

the-infinity:
Oh, good point: fixed all the site namings. Some of them made some sense when I started developing it, as Friedrichshafen uses a ParkingSite model for ParkingSpots, but in the end I decided to make a full transformation to spots, so your remark is absolutely correct.

AbdullahiFatola (approved):

PR #203, opened 2025-04-14
Integrate ulm_sensors for parking spots and parking sites

AbdullahiFatola:
This PR integrates the static and sensor parking data of the E-Quartiershub from the city of ULM. The data contains both realtime and static parking data mapped to `ParkingSites` and `ParkingSpots`.


AbdullahiFatola:
@the-infinity this PR depends on https://github.com/ParkenDD/parkapi-static-data/pull/8 That is the reason why the lint test partially failed.

tests/converters/kienzler_geojson_test.py
```diff
@@ -108,8 +108,8 @@ def test_get_static_parking_sites_with_geojson(
         assert static_parking_site_input.max_height == 1250
         assert static_parking_site_input.max_width == 800
         assert static_parking_site_input.park_and_ride_type == [ParkAndRideType.TRAIN]
-        assert static_parking_site_input.external_identifiers[0].type == ExternalIdentifierType.DHID
-        assert static_parking_site_input.external_identifiers[0].value == 'de:08317:14500_Parent'
+        assert static_parking_site_input.external_identifiers[0]['type'] == ExternalIdentifierType.DHID
```
the-infinity, line 111: Hmm. Why did you change this? Because, this would mean that the data is invalid. `static_parking_site_input.external_identifiers` should be a list of `ExternalIdentifierInput`, which is a (vali)dataclass. If you access it with `['type']`, you most likely have a dict, not a dataclass. This would crash the saving mechanism in ParkAPI, would make the typing at, and would also be a breaking change, which we should prevent. So the task is to find out why there are suddenly dicts instead if `ExternalIdentifierInput`, not changing the expected test results.

src/parkapi_sources/models/geojson_inputs.py
```diff
@@ -67,12 +88,20 @@ def to_static_parking_site_input(self, **kwargs) -> StaticParkingSiteInput:
...
+            **self.properties.to_dict(**kwargs),
+        )
+
     def update_static_parking_site_input(self, static_parking_site: StaticParkingSiteInput) -> None:
         static_parking_site.lat = self.geometry.coordinates[1]
         static_parking_site.lon = self.geometry.coordinates[0]
 
         for key, value in self.properties.to_dict().items():
-            setattr(static_parking_site, key, value)
+            if value is not None:
```
the-infinity, line 103: This means that you will never be able to remove fields. Eg if you have set a value at the GeoJSON and remove it later, then this removal will never reach ParkAPI, as you filter it out by this line. Is this by purpose? If yes, fine, just wanted to point out what's the effect of this change.

src/parkapi_sources/converters/ulm_sensors/validators.py
```diff
@@ -0,0 +1,67 @@
...
+    currentcarparkfullwithreservation: int | None = Noneable(IntegerValidator(allow_strings=True)), Default(None)
+    currentshorttermparker: int | None = Noneable(IntegerValidator(allow_strings=True)), Default(None)
+    timestamp: datetime = DateTimeValidator(
+        local_timezone=ZoneInfo('Europe/Berlin'),
+        target_timezone=timezone.utc,
+    )
+
+    def to_realtime_parking_site_input(self) -> RealtimeParkingSiteInput:
+        return RealtimeParkingSiteInput(
+            uid=str(self.id),
```
the-infinity, line 41: Why casting to string here?

src/parkapi_sources/models/geojson_inputs.py
```diff
@@ -39,13 +42,31 @@ def to_dict(self, *args, static_data_updated_at: datetime | None = None, **kwarg
...
+    address: Optional[str] = StringValidator(max_length=512), Default(None)
     description: Optional[str] = StringValidator(max_length=512), Default(None)
-    capacity: int = IntegerValidator()
-    has_realtime_data: bool = BooleanValidator()
+    capacity: Optional[int] = IntegerValidator(), Default(None)
+    has_realtime_data: Optional[bool] = BooleanValidator(), Default(None)
+    max_height: Optional[int] = Noneable(IntegerValidator()), Default(None)
+    max_width: Optional[int] = IntegerValidator(), Default(None)
+    park_and_ride_type: Optional[list[ParkAndRideType]] = ListValidator(EnumValidator(ParkAndRideType)), Default(None)
+    external_identifiers: Optional[list[ExternalIdentifierInput]] = (
```
the-infinity, line 55: You used modern typing (`int | None` instead of `Optional[int]`) above. I would be fine if you would use the old approach consistently, but mixing it is a bit strange, so I would suggest switching to modern typing without Optional everywhere.

docs/mapping/ulm_sensors.md
```diff
@@ -0,0 +1,37 @@
...
+|-------------------------------|----------------------------------------------|-------------|-----------------------------------------|------------------------------------------------------------------------------------|
+| id                            | string                                       | 1           | uid                                     |                                                                                    |
+| maxcarparkfull                | integer                                      | ?           | realtime_capacity                       |                                                                                    |
+| currentcarparkfulltotal       | integer                                      | ?           | realtime_free_capacity                  |                                                                                    |
+| timestamp                     | datetime                                     | ?           | realtime_data_updated_at                |                                                                                    |
```
the-infinity, line 20: In your validator, you set a cardinality of 1., here it's ?. One place has to be fixed :)

tests/converters/kienzler_geojson_test.py
AbdullahiFatola: It has now been updated to `ExternalIdentifierInput` instead of dict

src/parkapi_sources/models/geojson_inputs.py
AbdullahiFatola: I set it that way because some sources like `kienzler` have required fields that are missing in their `parkapi-static-data` e.g. `name` and this prevents the whole data from being processed by the converter, leading to `field_errors`. The "name" field is however present in the realtime data and is updated from the there.

src/parkapi_sources/converters/ulm_sensors/validators.py
AbdullahiFatola: I thought it would lead to ValidationError if the data source provides an integer, but I think it makes sense not the cast to string, since the ids are already alphanumeric from the source.

src/parkapi_sources/models/geojson_inputs.py
AbdullahiFatola: You're right i have now switched to modern typing.

docs/mapping/ulm_sensors.md
AbdullahiFatola: Ohhh that's right!

src/parkapi_sources/converters/ulm_sensors/validators.py
the-infinity: This would lead to an error at the validator: `id: str = StringValidator(min_length=1, max_length=256)` - this would already reject integers, so the casting would never happen anyway.

tests/converters/kienzler_geojson_test.py
the-infinity: Had a deeper look into that and fixed it in https://github.com/ParkenDD/parkapi-sources-v3/pull/203/commits/031897546294116708d9f0480c4f2bb80b7f3d63 .

src/parkapi_sources/models/geojson_inputs.py
the-infinity: Sadly, this approach was not that successful, you introduced the issue with `ExternalIdentifierInput`. Fixed it in the commit https://github.com/ParkenDD/parkapi-sources-v3/commit/031897546294116708d9f0480c4f2bb80b7f3d63 and adapted your `None` check.

I'm still not sure if this is a good idea, but as there is a better patch mechanisms then the GeoJSON one now, I'm fine with it.

PR #206, opened 2025-04-20
freiburg disabled

the-infinity:
Mapping definition for Freiburg disabled sources

### DIFF
--- a/docs/mapping/freiburg_disabled_sensors.md
+++ b/docs/mapping/freiburg_disabled_sensors.md
@@ -1,0 +1,19 @@
+# Freiburg disabled parking spots with sensors
+
+Freiburg provides a GeoJSON with sensor-enabled parking spots.
+
+## Properties
+
+| Field     | Type              | Cardinality | Mapping         | Comment |
+|-----------|-------------------|-------------|-----------------|---------|
+| name      | string            | 1           | uid, name       |         |
+| adresse   | hausnummer        | 1           | address         |         |
+| status    | [Status](#Status) | 1           | realtime_status |         |
+
+
+### Status
+
+| Key | Mapping   |
+|-----|-----------|
+| 0   | TAKEN     |
+| 1   | AVAILABLE |
--- a/docs/mapping/freiburg_disabled_static.md
+++ b/docs/mapping/freiburg_disabled_static.md
@@ -1,0 +1,14 @@
+# Freiburg static disabled parking spots
+
+Freiburg provides a GeoJSON with static parking spots.
+
+## Properties
+
+| Field      | Type       | Cardinality | Mapping     | Comment |
+|------------|------------|-------------|-------------|---------|
+| fid        | string     | 1           | uid         |         |
+| strasse    | hausnummer | 1           | address     |         |
+| hausnummer | string     | 1           | address     |         |
+| anzahl     | string     | 1           |             |         |
+| hinweis    | string     | 1           | description |         |
+| stadtteil  | string     | 1           |             |         |


docs/mapping/freiburg_disabled_static.md
```diff
@@ -0,0 +1,14 @@
...
+Freiburg provides a GeoJSON with static parking spots.
+
+## Properties
+
+| Field      | Type   | Cardinality | Mapping     | Comment |
+|------------|--------|-------------|-------------|---------|
+| fid        | string | 1           | uid         |         |
+| strasse    | string | 1           | address     |         |
+| hausnummer | string | 1           | address     |         |
+| anzahl     | string | 1           |             |         |
```
ThorstenFroehlinghaus, line 12: So, we will not map `anzahl`, which can be up to 12? This is ok for me - see also our BFRK discussion.
I suggest, to map `anzahl` != 1 as `description`.

docs/mapping/freiburg_disabled_static.md
```diff
@@ -0,0 +1,14 @@
+# Freiburg static disabled parking spots
+
+Freiburg provides a GeoJSON with static parking spots.
```
ThorstenFroehlinghaus, line 3: We will map the polygons to points, right? This should be mentions here.

docs/mapping/freiburg_disabled_sensors.md
```diff
@@ -0,0 +1,19 @@
+# Freiburg disabled parking spots with sensors
+
+Freiburg provides a GeoJSON with sensor-enabled parking spots.
+
+## Properties
+
+| Field     | Type              | Cardinality | Mapping         | Comment |
+|-----------|-------------------|-------------|-----------------|---------|
+| name      | string            | 1           | uid, name       |         |
+| adresse   | string            | 1           | address         |         |
```
ThorstenFroehlinghaus, line 10: Cardinality is 0..1, because `adresse` is sometimes empty.

docs/mapping/freiburg_disabled_sensors.md
```diff
@@ -0,0 +1,19 @@
+# Freiburg disabled parking spots with sensors
+
+Freiburg provides a GeoJSON with sensor-enabled parking spots.
+
+## Properties
+
+| Field     | Type              | Cardinality | Mapping         | Comment |
+|-----------|-------------------|-------------|-----------------|---------|
+| name      | string            | 1           | uid, name       |         |
```
ThorstenFroehlinghaus, line 9: I suggest, that `uid` contains only the prefix like `P01`.

docs/mapping/freiburg_disabled_sensors.md
ThorstenFroehlinghaus, line 10: The ending `, Germany` shall be removed from `adresse`.

docs/mapping/freiburg_disabled_sensors.md
```diff
@@ -0,0 +1,19 @@
...
+| adresse   | string            | 1           | address         |         |
+| status    | [Status](#Status) | 1           | realtime_status |         |
+
+
+### Status
+
+| Key | Mapping   |
+|-----|-----------|
+| 0   | TAKEN     |
```
ThorstenFroehlinghaus, line 18: It's just the other way: 0 = AVAILABLE, 1 = TAKEN.

ThorstenFroehlinghaus (changes requested):

docs/mapping/freiburg_disabled_static.md
the-infinity: In this case, it's still separate ParkingSpots, you will always find the other entries. So it's an 1:1 mapping, and we ignore the anzahl, as it's pretty visible.

docs/mapping/freiburg_disabled_static.md
the-infinity: Done (polygon is still preserved).

docs/mapping/freiburg_disabled_sensors.md
the-infinity: Emptystring is still a value of type string, so the cardinality is still 1. Added a comment about the emptystring mapping.

docs/mapping/freiburg_disabled_sensors.md
the-infinity: Documented

ThorstenFroehlinghaus (approved):

ThorstenFroehlinghaus (approved):

PR #215, opened 2025-05-13
Add freiburg static and dynamic park ride

AbdullahiFatola:
This PR adds the static and dynamic park and ride data from Freiburg GIS-System


tests/converters/freiburg_test.py
```diff
@@ -22,6 +26,37 @@ def freiburg_pull_converter(
     return FreiburgPullConverter(config_helper=mocked_static_geojson_config_helper, request_helper=request_helper)
 
 
+@pytest.fixture
```
the-infinity, line 29: As this is a different converter, please add it to a different test file.

src/parkapi_sources/converters/freiburg/models.py
```diff
@@ -31,6 +33,44 @@ class FreiburgPropertiesInput:
...
 
 
+class FreiburgParkingSiteTypeInput(Enum):
+    PARKPLATZ = 'Parkplatz'
+    PARKHAUS = 'Parkhaus'
+    TIEFGARAGE = 'Tiefgarage'
+    PARKNRIDE = 'Park&Ride'
+
+    def to_parking_site_type_input(self) -> ParkingSiteType:
+        # TODO: find out more details about this enumeration for a proper mapping
```
the-infinity, line 43: Is there a remaining TODO?

docs/mapping/freiburg.md
```diff
@@ -0,0 +1,74 @@
...
+| park_id                    | integer                  | 1           | uid                             |                                                                     |
+| name                       | string                   | 1           | name                            |                                                                     |
+| obs_max                    | integer                  | ?           | realtime_capacity/capacity      |                                                                     |
+| obs_free                   | integer                  | ?           | realtime_free_capacity          |                                                                     |
+| obs_ts                     | datetime                 | 1           | realtime_data_updated_at        |                                                                     |
+| obs_state                  | integer                  | ?           | #RealtimeOpeningStatus          |                                                                     |
```
the-infinity, line 52: Please add a link, and make an `RealtimeOpeningStatus` type based on that. The mapping is `realtime_opening_status`.

Also: The cardinality in your validator is 1, not ?. What is the correct value?

docs/mapping/freiburg.md
```diff
@@ -0,0 +1,74 @@
...
+* `purpose` is always set to `CAR`
+* `has_realtime_data` is always set to `True`
+
+A `ParkingSites` provides static and realtime data for a Park and Ride `ParkingSite`.
+
+| Field                      | Type                     | Cardinality | Mapping                         | Comment                                                             |
+|----------------------------|--------------------------|-------------|---------------------------------|---------------------------------------------------------------------|
+| park_id                    | integer                  | 1           | uid                             |                                                                     |
+| name                       | string                   | 1           | name                            |                                                                     |
+| obs_max                    | integer                  | ?           | realtime_capacity/capacity      |                                                                     |
```
the-infinity, line 49: `obs_max` and `obs_free` are both set to a cardinality of 1 in your validator, not ?. What's the correct value?

docs/mapping/freiburg.md
```diff
@@ -0,0 +1,74 @@
+# Realtime and Static data of the city of Freiburg
```
the-infinity, line 1: I do wonder if we could deduplicate the datasets right away. The realtime datasets are a subset of the static datasets, so the realtime data source will create duplicates of each and every entry in this dataset. As Freiburg seems to be very open about such things: I would suggest asking them if the realtime dataset could include a reference to the corresponding static dataset, so we can just create one source, where some of the datasets have realtime data, and some not.

docs/mapping/freiburg.md
```diff
@@ -0,0 +1,74 @@
...
+|----------------------------|--------------------------|-------------|---------------------------------|---------------------------------------------------------------------|
+| ogc_fid                    | integer                  | 1           | uid                             |                                                                     |
+| kapazitaet                 | integer                  | 1           | capacity                        |                                                                     |
+| name                       | string                   | 1           | name                            |                                                                     |
+| nummer                     | string                   | ?           | name                            |                                                                     |
+| kategorie                  | string                   | 1           | #ParkingSiteType                |                                                                     |
```
the-infinity, line 32: The Type is a enumeration `ParkingSiteType` which should be linked. The mapping is to `type`.

docs/mapping/freiburg.md
```diff
@@ -0,0 +1,74 @@
...
+* `has_realtime_data` is always set to `False`
+
+A `ParkingSites` provides static data for a Park and Ride `ParkingSite`.
+
+| Field                      | Type                     | Cardinality | Mapping                         | Comment                                                             |
+|----------------------------|--------------------------|-------------|---------------------------------|---------------------------------------------------------------------|
+| ogc_fid                    | integer                  | 1           | uid                             |                                                                     |
+| kapazitaet                 | integer                  | 1           | capacity                        |                                                                     |
+| name                       | string                   | 1           | name                            |                                                                     |
+| nummer                     | string                   | ?           | name                            |                                                                     |
```
the-infinity, line 31: `cardinality` is actually 1, as the field always exists and is always the correct type string, it's just emptystring. You can add that to the comment section that you will ignore emptystring.

tests/converters/freiburg_test.py
AbdullahiFatola: I have now created a separate test file for this.

docs/mapping/freiburg.md
AbdullahiFatola: They are now set to 1

docs/mapping/freiburg.md
AbdullahiFatola: We already asked the city of Freiburg and they said the data should be treated separately and not merged together. The 4 realtime parking sites come from their sensors.

the-infinity (approved):
LGTM. Before merging, you will have to rebase, though.

PR #219, opened 2025-05-19
Freiburg scanner mapping

the-infinity:
Freiburg provides data from scan cars

### DIFF
--- a/docs/mapping/freiburg_scanner.md
+++ b/docs/mapping/freiburg_scanner.md
@@ -1,6 +1,6 @@
 # Freiburg Scanner
 
-Freiburg Mercedes is a large dataset with scan car data.
+Freiburg Scanner is a large dataset with scan car data.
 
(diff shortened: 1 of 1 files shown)


docs/mapping/freiburg_scanner.md
```diff
@@ -0,0 +1,16 @@
+# Freiburg Scanner
+
+Freiburg Scanner is a large dataset with scan car data.
+
+Attributes which are set statically:
+
+* `has_realtime_data` is always set to `False`
+* `type` is always set to `ON_STREET`
```
ThorstenFroehlinghaus, line 8: * `purpose` is set to `CAR`
* `static_data_updated_at` is set to the moment of import

ThorstenFroehlinghaus (changes requested):
document purpose and static_data_updated_at

PR #223, opened 2025-05-19
ulm disabled mapping

the-infinity:
Proposal for Ulm disabled mapping

### DIFF
--- a/docs/mapping/ulm_disabled.md
+++ b/docs/mapping/ulm_disabled.md
@@ -6,6 +6,8 @@ Ulm provides a simple GeoJSON with Point geometry for disabled parking spots.
 * `static_data_updated_at` is set to the moment of import
 * `purpose` is set to `CAR`
 * `restricted_to.type` is set to `DISABLED`
+* `uid` is set to `{lat}_{lon}`
+* `lat` and `lon` are cut off 7 digits after the comma
 
 
 | Field    | Type         | Cardinality | Mapping   | Comment                                                                               |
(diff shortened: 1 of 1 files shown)


ThorstenFroehlinghaus (changes requested):
I suggest to use `lat`_`lon` as `id`, and to document this in the mapping.

docs/mapping/ulm_disabled.md
```diff
@@ -0,0 +1,14 @@
+# Ulm Disabled
+
+Ulm provides a simple GeoJSON with Point geometry for disabled parking spots.
+
+* `has_realtime_data` is set to `false`
+* `static_data_updated_at` is set to the moment of import
+* `purpose` is set to `CAR`
+* `restricted_to.type` is set to `DISABLED`
+
```
ThorstenFroehlinghaus, line 9: set `type` to `ON_STREET`

docs/mapping/ulm_disabled.md
```diff
@@ -0,0 +1,14 @@
...
+
+| Field    | Type         | Cardinality | Mapping   | Comment                                                                               |
+|----------|--------------|-------------|-----------|---------------------------------------------------------------------------------------|
+| image    | string (url) | 1           | photo_url |                                                                                       |
```
ThorstenFroehlinghaus, line 13: use lat_lon as id
will we reduce the precision of lat/lon to 6 or 7 digits?

docs/mapping/ulm_disabled.md
the-infinity: Would not recommend this: there are substantial amounts of parking spots in off street parking areas.

docs/mapping/ulm_disabled.md
the-infinity: Updated

docs/mapping/ulm_disabled.md
ThorstenFroehlinghaus: Ok, I agree

PR #224, opened 2025-05-19
karlsruhe disabled mapping

the-infinity:
Karlsruhe disabled mapping proposal

### DIFF
--- a/docs/mapping/karlsruhe_disabled.md
+++ b/docs/mapping/karlsruhe_disabled.md
@@ -12,8 +12,8 @@ Karlsruhe provides a GeoJSON with Point geometry for disabled parking spots.
 | gemeinde      | string             | 1           | address                | `address` will be `{stadtteil}, {standort}`                                               |
 | stadtteil     | string             | 1           |                        |                                                                                           |
 | standort      | string             | 1           | name, address          |                                                                                           |
-| parkzeit      | integer            | ?           | description            |                                                                                           |
-| max_parkdauer | string             | ?           | description            |                                                                                           |
+| parkzeit      | string             | ?           | description            | Unclear format, no reliable converter to `restricted_to.hours` possible.                  |
+| max_parkdauer | string             | ?           | description            | Unclear format, no reliable converter to `restricted_to.may_stay` possible.               |
 | stellplaetze  | integer            | 1           |                        | If stellplaetze is more then 1, multiple slightly distributed `ParkingSpot`s are created. |
 | bemerkung     | strong             | 1           | description            |                                                                                           |
 | stand         | string (date-time) | 1           | static_data_updated_at |                                                                                           |
--- a/docs/mapping/karlsruhe_disabled.md
+++ b/docs/mapping/karlsruhe_disabled.md
@@ -1,6 +1,6 @@
-# Ulm Disabled
+# Karlsruhe Disabled
 
-Ulm provides a GeoJSON with Point geometry for disabled parking spots.
+Karlsruhe provides a GeoJSON with Point geometry for disabled parking spots.
 
 * `has_realtime_data` is set to `false`
 * `purpose` is set to `CAR`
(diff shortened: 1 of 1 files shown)


docs/mapping/karlsruhe_disabled.md
```diff
@@ -0,0 +1,19 @@
+# Ulm Disabled
```
ThorstenFroehlinghaus, line 1: # Karlsruhe Disabled

docs/mapping/karlsruhe_disabled.md
```diff
@@ -0,0 +1,19 @@
...
+
+Ulm provides a GeoJSON with Point geometry for disabled parking spots.
```
ThorstenFroehlinghaus, line 3: Karlsruhe ...
Should we mention, that this is not a static file, but data retrieved via an URL?
Should we mention the retrieval frequency?

docs/mapping/karlsruhe_disabled.md
```diff
@@ -0,0 +1,19 @@
...
+
+* `has_realtime_data` is set to `false`
+* `purpose` is set to `CAR`
+* `restricted_to.type` is set to `DISABLED`
+
+| Field         | Type               | Cardinality | Mapping                | Comment                                                                                   |
+|---------------|--------------------|-------------|------------------------|-------------------------------------------------------------------------------------------|
+| id            | integer            | 1           | uid                    |                                                                                           |
```
ThorstenFroehlinghaus, line 11: `id` is a string, like b"ehinderten_parkplaetze.1"

docs/mapping/karlsruhe_disabled.md
```diff
@@ -0,0 +1,19 @@
...
+| gemeinde      | string             | 1           | address                |                                                                                           |
+| stadtteil     | string             | 1           | address                |                                                                                           |
+| standort      | string             | 1           | name                   |                                                                                           |
+| parkzeit      | integer            | ?           | description            |                                                                                           |
```
ThorstenFroehlinghaus, line 15: `parkzeit` is a string, like "täglich 8 bis 20 Uhr"
We should mention, that `parkzeit` could be mapped to `restricted_to.hours`, but developing such a logic is too much effort.

docs/mapping/karlsruhe_disabled.md
```diff
@@ -0,0 +1,19 @@
...
+| max_parkdauer | string             | ?           | description            |                                                                                           |
```
ThorstenFroehlinghaus, line 16: Can be mapped to `restricted_to.may_stay`, or? We should look for patterns like "x Std" or "y min".

docs/mapping/karlsruhe_disabled.md
```diff
@@ -0,0 +1,19 @@
...
+
+* `has_realtime_data` is set to `false`
+* `purpose` is set to `CAR`
+* `restricted_to.type` is set to `DISABLED`
+
+| Field         | Type               | Cardinality | Mapping                | Comment                                                                                   |
+|---------------|--------------------|-------------|------------------------|-------------------------------------------------------------------------------------------|
+| id            | integer            | 1           | uid                    |                                                                                           |
+| gemeinde      | string             | 1           | address                |                                                                                           |
+| stadtteil     | string             | 1           | address                |                                                                                           |
```
ThorstenFroehlinghaus, line 13: not needed for `address`

docs/mapping/karlsruhe_disabled.md
```diff
@@ -0,0 +1,19 @@
...
+| standort      | string             | 1           | name                   |                                                                                           |
```
ThorstenFroehlinghaus, line 14: Also mapped to `address`. We should mention the pattern `address`= `standort` + ", " + `gemeinde`

ThorstenFroehlinghaus (changes requested):

docs/mapping/karlsruhe_disabled.md
the-infinity: See `parkzeit`

docs/mapping/karlsruhe_disabled.md
ThorstenFroehlinghaus: I agree, we should just mention in the mapping file, that we do not derive `retricted_to.hours` because it is too much effort.

docs/mapping/karlsruhe_disabled.md
ThorstenFroehlinghaus: Ok, but we should document why we don't derive `restricted_to.may_stay`: too much effort, only few datasets.

ThorstenFroehlinghaus (approved):
LGTM

PR #226, opened 2025-05-21
Smaller Freiburg Disabled Enhancements

the-infinity:
Adds `DISABLED` restriction to both Freiburg disabled converters, and locality name to the static one. Adds more docs about static values.


ThorstenFroehlinghaus (approved):
LGTM

ThorstenFroehlinghaus (approved):
LGTM

PR #237, opened 2025-07-08
Integrate B+B Parkhaus Push Converter

AbdullahiFatola:
This PR integrates the static for data Parking sites from the B+B Parkhaus GmbH. It is a push converter for excel sheet.


src/parkapi_sources/converters/bb_parkhaus/converter.py
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # For some reason, BB Parkhaus decided to change the titles
+        bb_parkhaus_header_rows: dict[str, str] = {
+            'Zweck der Anlage': 'purpose',
+            'Ladeplätze': 'capacity_charging',
+            'Gebührenpflichtig?': 'has_fee',
+            'is_covered': 'is_covered',
```
the-infinity, line 46: That's the same. I guess you meant something else?

docs/mapping/bb_parkhaus.md
```diff
@@ -0,0 +1,64 @@
+# B+B Parkhaus GmbH & Co. KG
+
+B+B Parkhaus GmbH & Co. KG provides an Excel table with some static parking site data.
+
+Static values:
+
+* `purpose` is always `CAR`
+* `has_realtime_data` is always `false`
```
the-infinity, line 8: `static_data_updated_at` definition is missing

src/parkapi_sources/converters/bb_parkhaus/converter.py
AbdullahiFatola: In the Excel sheet, the column is named `is_covered` . If i do not added to the `bb_parkhaus_header_rows``, then it have an `UnsetValue` result. That's why I included it in there.

PR #243, opened 2025-07-14
freiburg_disabled_sensors: Set type=ON_STREET

AbdullahiFatola:
This PR sets `type` as `ON_STREET` in `freiburg_disabled_sensors` and `freiburg_disabled_static`.

### DIFF
--- a/docs/mapping/freiburg_disabled_static.md
+++ b/docs/mapping/freiburg_disabled_static.md
@@ -7,7 +7,6 @@ Freiburg provides a GeoJSON with static parking spots. lat / lon are set to the 
 * `purpose` is set to `CAR`
 * `has_realtime_data` is set to `false`
 * `static_data_updated_at` is set to the moment of import
-* `type` is set to `ON_STREET`
 
 ## Properties
 


src/parkapi_sources/converters/freiburg_disabled_sensors/models.py
```diff
@@ -48,6 +48,7 @@ def to_static_parking_spot_input(self) -> StaticParkingSpotInput:
             static_data_updated_at=datetime.now(tz=timezone.utc),
             lat=self.geometry.coordinates[1],
             lon=self.geometry.coordinates[0],
+            type=ParkingSpotType.ON_STREET,
```
the-infinity, line 51: This is almost never true: over 50 % (maybe even all?) of the parking spots are different type of parking garages, not on street.

src/parkapi_sources/converters/freiburg_disabled_static/models.py
```diff
@@ -56,6 +56,7 @@ def to_static_parking_spot_input(self) -> StaticParkingSpotInput:
             static_data_updated_at=datetime.now(tz=timezone.utc),
             lat=round_7d(Decimal(point.y)),
             lon=round_7d(Decimal(point.x)),
+            type=ParkingSpotType.ON_STREET,
```
the-infinity, line 59: This is also not valid for several datasets: I do see P+R Parkplatz, Parkplatz Reiterstrasse or Parkplatz Ecke Hildastr right away, and have several other datasets which might have the same issue. Still do that?

AbdullahiFatola:
> Also, this should be added to the document if actually a thing.

It is now added to the docs

src/parkapi_sources/converters/freiburg_disabled_sensors/models.py
ThorstenFroehlinghaus: Freiburg told us: "die sind alle ON_STREET (wobei teilweise zufahrtsbeschränkt mit „Anlieger“-Schild)". We told them about our doubts, but they confirmed their statement.

PR #244, opened 2025-07-18
Extend Friedrichshafen

the-infinity:
Adds better datex2 mapping because Friedrichshafen wants to use that


docs/mapping/friedrichshafen_sensors.md
```diff
@@ -15,18 +15,18 @@ Attributes which are set statically:
...
+| version                    | integer                                               | 1           |                        |                                                         |
+| parkingName                | string                                                | 1           | name                   |                                                         |
+| parkingRecordVersionTime   | string (datetime)                                     | 1           | static_data_updated_at |                                                         |
+| parkingNumberOfSpaces      | integer                                               | 1           |                        | Has to be 1 at all times, will be validated             |
+| parkingLocation            | [PointByCoordinates](#PointByCoordinates)             | 1           |                        |                                                         |
+| assignedParkingAmongOthers | ApplicableForUser                                     | 1           | restricted_to.type     |                                                         |
```
ThorstenFroehlinghaus, line 26: [ApplicableForUser](#ApplicableForUser)

docs/mapping/friedrichshafen_sensors.md
```diff
@@ -15,18 +15,18 @@ Attributes which are set statically:
...
+| parkingLayout              | [ParkingLayoutEnum](#ParkingLayoutEnum)               | 1           | type                   |                                                         |
```
ThorstenFroehlinghaus, line 27: Should parkingLayout be optional? If not, which value do we choose for on-street parking?

docs/mapping/friedrichshafen_sensors.md
```diff
@@ -36,6 +36,33 @@ A `ParkingRecord` provides static data for a `ParkingSpot`.
...
+### ParkingLayoutEnum
+
+| Key                       | Mapping                   |
+|---------------------------|---------------------------|
+| multiStorey               | CAR_PARK                  |
+| singleLevel               | OFF_STREET_PARKING_GROUND |
+| underground               | UNDERGROUND               |
+| undergroundAndMultiStorey | CAR_PARK                  |
+| automatedParkingGarage    | CAR_PARK                  |
+| openSpace                 |                           |
```
ThorstenFroehlinghaus, line 48: no mapping for `openSpace`?

ThorstenFroehlinghaus (changes requested):
We need to define, how parkingLayout shall be set for on-street parking

docs/mapping/friedrichshafen_sensors.md
the-infinity: As `openSpace` means `A normal ground level parking place.`, there is no clear way to determine if it's `OFF_STREET_PARKING_GROUND` or `ON_STREET`. This gets specified by `urbanParkingSiteType`, though.

docs/mapping/friedrichshafen_sensors.md
the-infinity: It is accoring to Datex2 spec.

ThorstenFroehlinghaus (approved):
LGTM

minor issue: parkingLayout cardinality should be `?` instead of `*`, right?

the-infinity:
@ThorstenFroehlinghaus actually, the standard says * . Changed the handling here: https://github.com/ParkenDD/parkapi-sources-v3/pull/245/files

PR #258, opened 2025-08-07
Integrate parking sites for city of Aalen

AbdullahiFatola:
This PR integrates the static and realtime parking sites data of the Stadtwerke Aalen GmbH.


README.md
```diff
@@ -66,6 +124,7 @@ We support following data sources:
 | Verband Region Stuttgart: Neustadt                             | ParkingSite | car        | pull         | `vrs_neustadt`              | yes      |
 | Verband Region Stuttgart: Park and Ride                        | ParkingSite | car        | push (xlsx)  | `vrs_p_r`                   | no       |
 | Verband Region Stuttgart: Vaihingen                            | ParkingSite | car        | pull         | `vrs_vaihingen`             | yes      |
+>>>>>>> 5bd129be1f80fac5f8ebf139c3d328ac28c24d88
```
the-infinity, line 127: You missed some rebase artefacts

src/parkapi_sources/converters/aalen/converter.py
```diff
@@ -0,0 +1,95 @@
...
+
+        realtime_aalen_inputs, import_realtime_parking_site_exceptions = self._get_raw_realtime_parking_sites()
+        import_parking_site_exceptions += import_realtime_parking_site_exceptions
+
+        static_parking_site_inputs_by_uid: dict[str, StaticParkingSiteInput] = {}
+        for static_parking_site_input in static_parking_site_inputs:
+            static_parking_site_inputs_by_uid[static_parking_site_input.uid] = static_parking_site_input
+
+        for realtime_aalen_input in realtime_aalen_inputs:
+            # If the uid is not known in our static data: ignore the realtime data
```
the-infinity, line 51: This is a quite complicated approach for matching. Why don't you just use the realtime data source name as `uid`? `P1 Rathaus` is a perfectly fine uid.

Has two advantages:
- taking less ressources
- less likely to run into rate limits from Github


And it's no problem to have more realtime datasets then static ones, this happens at other converters, too. Adds an error to the counter, but is handled well, without noisy log entries: https://github.com/ParkenDD/park-api-v3/blob/main/webapp/services/import_service/generic/generic_parking_site_import_service.py#L142

src/parkapi_sources/converters/aalen/models.py
```diff
@@ -0,0 +1,47 @@
...
+            self.OCCUPIED: OpeningStatus.OPEN,
+            self.CLOSED: OpeningStatus.CLOSED,
+        }.get(self)
+
+
+@validataclass
+class AalenInput:
+    name: str = StringValidator()
+    status: AalenOpeningStatus = EnumValidator(AalenOpeningStatus)
+    max: int = IntegerValidator(allow_strings=True)
```
the-infinity, line 33: Why do you allow strings here?

README.md
AbdullahiFatola: oh...i have now removed them.

src/parkapi_sources/converters/aalen/converter.py
AbdullahiFatola: That makes sense. I have updated the static data in this PR https://github.com/ParkenDD/parkapi-static-data/pull/17

PR #266, opened 2025-08-12
freiburg scanner address

the-infinity:
Uses the address which is provided by Freiburg Scanners now


ThorstenFroehlinghaus (changes requested):
Please update also https://github.com/ParkenDD/parkapi-sources-v3/blob/main/docs/mapping/freiburg_scanner.md

the-infinity:
Done

ThorstenFroehlinghaus (approved):
LGTM

PR #294, opened 2025-10-15
freiburg_disabled_sensors: Set type=ON_STREET

AbdullahiFatola:
This PR sets type as ON_STREET in freiburg_disabled_sensors. It references this PR https://github.com/ParkenDD/parkapi-sources-v3/pull/243 which was closed by mistake.


the-infinity:
It got a bit better, but there are still multiple non-ON_STREET parking spaces:

image
image
image

PR #331, opened 2025-12-29
fix herrenberg bike url

the-infinity:
Fixes url after url change


AbdullahiFatola (commented):
The URL also seems not to work again

AbdullahiFatola:
Now it works but still validation errors, the test was using old URL.

> When i run a test with `test-pull-converter.py` using the new URL, i get the following.
> 
> ```
> raise ImportSourceException(
> parkapi_sources.exceptions.ImportSourceException: ImportSourceException herrenberg_bike: Invalid Input at source herrenberg_bike: {'code': 'field_errors', 'field_errors': {'type': {'code': 'required_field'}, 'features': {'code': 'required_field'}}}, data: {'error': 'distribution not found'}
> ```

the-infinity:
Ohh, fun. Sent an mail with you in CC, @AbdullahiFatola

derhuerst (changes requested):
Until either Herrenberg or we change the data structure this won't work, so I'll mark this PR accordingly.

hbruch:
Deprecated. Replace by #349

PR #349, opened 2026-02-12
[PARK-1]: rewrite converters for herrenberg bikes

ninanomenon:
This merge request changes two things:
- It updates the data transformation according to: https://github.com/ParkenDD/parkapi-sources-v3/blob/main/docs/mapping/herrenberg_bike.md
- Since I am a typing nerd, I sprinkled some more typing information where I found a quick solution. In the long run this will help a lot.

### DIFF
--- a/src/parkapi_sources/converters/herrenberg_bike/models.py
+++ b/src/parkapi_sources/converters/herrenberg_bike/models.py
@@ -24,6 +24,7 @@ from parkapi_sources.models.shared_inputs import ExternalIdentifierInput
 from parkapi_sources.util import round_7d
 from parkapi_sources.util.dict import AnyDict
 from parkapi_sources.validators import MappedBooleanValidator
+from parkapi_sources.validators.string_validators import ReplacingStringValidator
 
 
 class HerrenbergBikeType(StrEnum):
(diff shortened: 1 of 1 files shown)


src/parkapi_sources/converters/herrenberg_bike/models.py
```diff
@@ -3,178 +3,130 @@
...
+                return ParkingSiteType.STANDS
+            case self.REINE_VORDERRADHALTERUNG:
+                return ParkingSiteType.WALL_LOOPS
+            case self.BÜGEL_PLUS_VORDERRADHALTERUNG:
+                return ParkingSiteType.SAFE_WALL_LOOPS
+            case self.ANDERE:
+                return ParkingSiteType.FLOOR
+
+
+class HerrenbergProperties(TypedDict):
```
JGaukrogers: I don't see lon/lat fields. Is that correct? I ask because according to [this](https://github.com/ParkenDD/parkapi-sources-v3/blob/main/docs/mapping/herrenberg_bike.md) document, there should be.

src/parkapi_sources/converters/herrenberg_bike/models.py
```diff
@@ -3,178 +3,130 @@
...
-            result['name'] = 'Fahrrad-Abstellanlagen'
-        if self.date_surveyed:
-            result['static_data_updated_at'] = datetime.combine(self.date_surveyed, time(), tzinfo=timezone.utc)
-        else:
-            result['static_data_updated_at'] = datetime.now(timezone.utc)
-        if self.supervision_type:
-            result['supervision_type'] = self.supervision_type.to_supervision_type()
-
-        return result
+    fid: int = IntegerValidator(min_value=1)
```
JGaukrogers: I don't see a field `coordinates` :O

PR #351, opened 2026-02-23
UV

ninanomenon:
Modern python package managment.


the-infinity:
Replaced by https://github.com/ParkenDD/parkapi-sources-v3/pull/392

PR #372, opened 2026-04-08
Mapping easypark friedrichshafen push converter

AbdullahiFatola:
This PR creates the mapping of the Onstreet parking sites from Friedrichshafen. This will be a push converter, since the data is static and provided in CSV.


docs/mapping/friedrichshafen_easypark.md
```diff
@@ -0,0 +1,63 @@
+# Friedrichshafen EasyPark
+
+Friedrichshafen provides a static CSV file of Onstreet Parking sites inventory. Therefore, this will be a push converter.
+
+Static values:
+
+* `name` is always `Straßenparkplatz`
+* `type` is always `PARKING_CENTER_LINE`
```
ThorstenFroehlinghaus: We should mention that this is a `LinearParkingSite` with a `LinearParkingSiteType`.

docs/mapping/friedrichshafen_easypark.md
```diff
@@ -0,0 +1,63 @@
...
+* `purpose` is always `CAR`
+* `has_realtime_data` is always `false`
+* Sites having the field `park_angle` with value `no_parking` should not be integrated.
+
+
+| Field                     | Type                 | Cardinality | Mapping                                                     | Comment                                                                       |
+|---------------------------|----------------------|-------------|-------------------------------------------------------------|-------------------------------------------------------------------------------|
+| id                        | integer              | 1           | uid                                                         |                                                                               |
+| length                    | integer              | 1           | capacity/[LinearParkingSiteSides](#LinearParkingSiteSides)  |                                                                               |
```
ThorstenFroehlinghaus: The unit of `length` needs to be clarified. These are very large numbers like 7,11523E+14.

docs/mapping/friedrichshafen_easypark.md
```diff
@@ -0,0 +1,63 @@
...
+| park_angle                | string               | 1           | [LinearParkingSiteSides](#LinearParkingSiteSides)           | Entries with `no_parking` should not be integrated.                           |
```
ThorstenFroehlinghaus: For enums we should define a source type like https://github.com/mobidata-bw/parkapi-sources-v3/blob/main/docs/mapping/bfrk_bw_car.md#BfrkBauart
The `Type` column should reference a source type `ParkAngle`, and the `Mapping` column should be `sides`[`side`=LinearParkingSiteSideType(`street_side`)].`orientation`.

docs/mapping/friedrichshafen_easypark.md
```diff
@@ -0,0 +1,63 @@
...
+| street_side               | string               | 1           | [LinearParkingSiteSides](#LinearParkingSiteSides)           |                                                                               |
```
ThorstenFroehlinghaus: The `Type` column should reference a source type `StreetSide`, and the `Mapping` column should be `sides`[`side`=LinearParkingSiteSideType(`street_side`)].`side`.

docs/mapping/friedrichshafen_easypark.md
```diff
@@ -0,0 +1,63 @@
...
+| location_on_sidewalk      | numeric              | 1           | lat/lon/geojson                                             | The center of the LineString coordinates is used as latitude and longitude    |
+| permissions_translation   | string               | 1           | description                                                 |                                                                               |
```
ThorstenFroehlinghaus: `has_fee`=true if `permissions`=`FreeParking`, else false.

ThorstenFroehlinghaus: `permissions_translation` should also be mapped to `fee_description`.

docs/mapping/friedrichshafen_easypark.md
```diff
@@ -0,0 +1,63 @@
...
+| perpendicular | PERPENDICULAR |
+| diagonal      | DIAGONAL      |
+| no_parking    |               |
+
+
+### ParkAngleCapacity
+
+| Key           | Mapping       | Comment                                              |
+|---------------|---------------|------------------------------------------------------|
+| parallel      | capacity      | Calculate `capacity` with `Math.round(length / 5)`   |
```
ThorstenFroehlinghaus: use `Math.ceil`

ThorstenFroehlinghaus (changes requested):
We need to clarify the notation for structured mappings, where, e.g., an element park_angle=parallel is mapped to sides[{side=RIGHT,orientation=PARALLEL}]

docs/mapping/friedrichshafen_easypark.md
AbdullahiFatola: Thanks for the hint, I will update it to reflect the `LinearParkingSiteType`,

docs/mapping/friedrichshafen_easypark.md
richardforsterNVBW: It is meters. If we have `parking_angle` == `parallel` AND `length` == 7.1, then `capacity` is 1 since `Math.round(7.1 / 5)`. The estimation is that for parallel parking only one car can fit in this parking lot.

docs/mapping/friedrichshafen_easypark.md
richardforsterNVBW: What if `permissions` == `FreeParking, PaidParking` or `FreeParking, ParkingMeter`?

docs/mapping/friedrichshafen_easypark.md
AbdullahiFatola: To avoid complexity, it is all defined in `#LinearParkingSiteSides`

docs/mapping/friedrichshafen_easypark.md
AbdullahiFatola: Also defined in `#LinearParkingSiteSides`

docs/mapping/friedrichshafen_easypark.md
AbdullahiFatola: That make sense to include `has_fee`, but the values of `permissions` is not consistent and has more comma-separated values in addition to `FreeParking`. Therefore, it can be misleading to set `has_fee` with values of `permissions`. 

Instead, I would set the `has_fee=true` if `permission_period` contains `Gebührenpflichtig`.

AbdullahiFatola: Now added!

docs/mapping/friedrichshafen_easypark.md
AbdullahiFatola: I would leave it open for the developer to decide, but add to comment that the result of capacity calculation should be a whole number.

docs/mapping/friedrichshafen_easypark.md
ThorstenFroehlinghaus: correction: `has_fee`=false if and only if `permission` is exactly `FreeParking`

AbdullahiFatola: I would avoid mapping `has_fee` using `permissions`, because mapping everything else as `true` will be misleading if  `permissions` is not `FreeParking`.

docs/mapping/friedrichshafen_easypark.md
richardforsterNVBW: I wouldn't use `Math.ceil` because that function always rounds up. In the current map implementation, which was developed in collaboration with the city of Friedrichshafen, it was agreed to round up or down depending on the nearest number. In this case, 1.2 is rounded down to 1 and 1.5 is rounded up to 2.

docs/mapping/friedrichshafen_easypark.md
richardforsterNVBW: I [updated the CSV](https://github.com/mobidata-bw/ipl-planning/blob/main/data/parking/Stadt_Friedrichshafen/EasyPark_Parkraumerhebung/260417_friedrichshafen_inventory.csv) with a correct length representation in meters.

docs/mapping/friedrichshafen_easypark.md
ThorstenFroehlinghaus: We should round down these fractional numbers to ensure these `capacity` defines the number of cars which surely can park here.

AbdullahiFatola: Now included!

docs/mapping/friedrichshafen_easypark.md
```diff
@@ -0,0 +1,69 @@
...
+| location_on_sidewalk      | numeric                     | 1           | lat/lon/geojson                                             | The center of the LineString coordinates is used as latitude and longitude                                                         |
+| permissions_translation   | string                      | 1           | description/fee_description                                 |                                                                                                                                    |
+| permission_period         | string                      | ?           | fee_description                                             |                                                                                                                                    |
+| time_limited              | string                      | ?           | fee_description                                             | If available, add to  `fee_description`                                                                                            |
+
+
+### LinearParkingSiteSide
```
the-infinity: I wonder if we should do that, because after defining it, it's the first dataset which actually uses sides. Is it expected to get more datasets with defined sides?

AbdullahiFatola: I defined it since the information on e.g. orientation, length are provided in the data. Are there other ways to define it?

AbdullahiFatola: @the-infinity would there be any problem when mapping, if we define sides, without more datasets with defined sides?

the-infinity: Would make the follow-up change in ParkAPI rather large, and it would create own views for the Geoserver. Therefore, much more complexity with ... well, limited usage?

richardforsterNVBW: Currently, the `orientation` can only be mapped to `sides`, where a left and a right `side` can be defined.
Since all the lines I have checked are assigned to a specific side of the road, `orientation` could also be mapped to the top level without distinguishing between the right and left sides of a line. This information is only required for `ROAD_CENTER_LINE`s. However, this would require a revision of the data model. Currently, such a mapping is not planned.

AbdullahiFatola: Ok. Since we still have limited usage of `sides` and to avoid complexity with the geoserver views, I have now removed the `LinearParkingSiteSide`  and its relations from the mapping document

AbdullahiFatola: @the-infinity any idea why the lint test fails, even though only the mapping doc was modified.

AbdullahiFatola: I found the problem. It seems you have updated the `konstanz_bike` to the new pull converter and the pytest for `konstanz_bike_test` seems to encounter an AssertionError

AbdullahiFatola: I will try to fix it here, so it can let the lint test for this MR pass through. 

@the-infinity ...however, before I proceed, is the lint test by chance pulling data in realtime from the Konstanz URL or using the `konstanz_bike.json` data in repository `test` folder. Because I am still asking myself how the lint test passed during your MR for the Konstanz converter? or am I missing something?

AbdullahiFatola: I see you fixed it already in this MR https://github.com/ParkenDD/parkapi-sources-v3/pull/376, thanks 


> I will try to fix it here, so it can let the lint test for this MR pass through.
> 
> @the-infinity ...however, before I proceed, is the lint test by chance pulling data in realtime from the Konstanz URL or using the `konstanz_bike.json` data in repository `test` folder. Because I am still asking myself how the lint test passed during your MR for the Konstanz converter? or am I missing something?

the-infinity (approved):

PR #378, opened 2026-05-26
enforce purpose

the-infinity:
enforces purpose

@AbdullahiFatola this is a kind of breaking MR, as it enforces a new value to be set. At mannheim and buchen as well as at any generic source in ParkAPI, I would add a compatibility layer, but this should not be active forever. Can you check all generic data sources and ask them to add purpose?

Also, we should move mannheim and buchen to a generic data source in the future, as they speak ParkAPI. The concept of a generic data source did not exist in early days, but now it does.

### DIFF
--- a/src/parkapi_sources/converters/aalen/converter.py
+++ b/src/parkapi_sources/converters/aalen/converter.py
@@ -11,7 +11,7 @@ from parkapi_sources.converters.base_converter.pull import (
     StaticGeojsonDataMixin,
 )
 from parkapi_sources.exceptions import ImportParkingSiteException
-from parkapi_sources.models import RealtimeParkingSiteInput, SourceInfo, StaticParkingSiteInput
+from parkapi_sources.models import PurposeType, RealtimeParkingSiteInput, SourceInfo, StaticParkingSiteInput
 
 from .models import AalenInput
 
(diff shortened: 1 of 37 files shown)


AbdullahiFatola (approved):
LGTM. We will ask Mannheim and Buchen to update their ParkingSite data with `purpose`, but I do not know how long this takes. Therefore, it makes sense to have the compatibility layer which you want to create, before it is finally enforced. I would also check if there are other sources affected in [ParkenDD/parkapi-static-data](https://github.com/ParkenDD/parkapi-static-data)  and others.

Plus, I think this enforcement only affects data integrated before `purpose` was introduced and after we introduced it, the [OpenAPI Documentation](https://api.mobidata-bw.de/park-api/documentation/public.html#/paths/v3-parking-sites/get) has always have the `purpose` as `required`, but seems not to be enforced in the models or base_converters.

the-infinity:
> LGTM. We will ask Mannheim and Buchen to update their ParkingSite data with `purpose`, but I do not know how long this takes. Therefore, it makes sense to have the compatibility layer which you want to create, before it is finally enforced. I would also check if there are other sources affected in [ParkenDD/parkapi-static-data](https://github.com/ParkenDD/parkapi-static-data) and others.

Mannheim and Buchen already got the compatibility layer. But all sources using the ` generic_converter` approach would not: https://github.com/ParkenDD/park-api-v3#source-configuration , eg Stuttgart (any maybe more, don't have access to ansible from vacation).

PR #382, opened 2026-06-05
Park 5 friedrichshafen easypark converter

JGaukrogers:
What I did:
* Add a converter for data for Friedrichshafen Easypark.
* Add a validator for geoJSON data coming from a csv file. It only supports LineStrings, but it should be easy to get it to support other coordinate types in the future.


src/parkapi_sources/converters/friedrichshafen_easypark/converter.py
```diff
@@ -0,0 +1,67 @@
...
+class FriedrichshafenEasyParkPushConverter(CsvConverter, ParkingSiteBaseConverter):
+    friedrichshafen_easypark_row_validator = DataclassValidator(FriedrichshafenEasyParkRowInput)
+
+    source_info = SourceInfo(
+        uid='friedrichshafen_easypark',
+        name='Friedrichshafen EasyPark',
+        has_realtime_data=False,
+    )
+
+    header_mapping: dict[str, str] = {
```
ninanomenon: This typehint is superfluous

src/parkapi_sources/converters/friedrichshafen_easypark/converter.py
```diff
@@ -0,0 +1,67 @@
...
+        static_parking_site_inputs: list[StaticParkingSiteInput] = []
+        static_parking_site_errors: list[ImportParkingSiteException] = []
+
+        mapping: dict[str, int] = self.get_mapping_by_header(self.header_mapping, data[0])
+
+        # We start at row 2, as the first one is our header
+        for row in data[1:]:
+            input_dict: dict[str, str] = {}
+            for field in self.header_mapping.values():
+                input_dict[field] = row[mapping[field]]
```
ninanomenon, line 49: Is there a reason why you do not use the csv parser from python?

tests/converters/friedrichshafen_easypark_test.py
```diff
@@ -0,0 +1,36 @@
...
+def friedrichshafen_easypark_push_converter(
+    mocked_config_helper: Mock, request_helper: RequestHelper
+) -> FriedrichshafenEasyParkPushConverter:
+    return FriedrichshafenEasyParkPushConverter(config_helper=mocked_config_helper, request_helper=request_helper)
+
+
+class FriedrichshafenEasyParkPushConverterTest:
+    @staticmethod
+    def test_get_static_parking_sites(friedrichshafen_easypark_push_converter: FriedrichshafenEasyParkPushConverter):
+        with get_data_path('friedrichshafen_inventory.csv').open(encoding='utf-8-sig') as friedrichshafen_easypark_file:
```
ninanomenon, line 26: What is `utf-8-sig`?

tests/converters/friedrichshafen_easypark_test.py
```diff
@@ -0,0 +1,36 @@
...
+            friedrichshafen_easypark_data = StringIO(friedrichshafen_easypark_file.read())
```
ninanomenon, line 27: What is `StringIO` doing here?

src/parkapi_sources/converters/friedrichshafen_easypark/converter.py
JGaukrogers: Deleted

src/parkapi_sources/converters/friedrichshafen_easypark/converter.py
JGaukrogers: This converter inherits from CsvConverter. `handle_csv` is called from handle_csv_string, where the csv parser is used o read the csv file and convert the content into a list of lists. Does that answer your question somewhat? 

I tried to find out if there is a better way to parse the content into `StaticParkingSiteInput` objects, but I couldn't fine anything yet.

tests/converters/friedrichshafen_easypark_test.py
JGaukrogers: It's the encoding we need here to avoid bad invisible characters messing around. Here is more context for this: [#mobility-data-bw > ParkAPI/Friedrichshafen EasyPark @ 💬](https://chat.letterbox-online.de/#narrow/channel/44-mobility-data-bw/topic/ParkAPI.2FFriedrichshafen.20EasyPark/near/248742)

tests/converters/friedrichshafen_easypark_test.py
ninanomenon: bump

JGaukrogers: `StringIO` lets you treat strings as file-like objects. The next thing we do is call `handle_csv_string` with the data we read previously, and then the data is read again with `csv.reader`. `csv.reader` seems to not to be able to handle strings, and with `StringIO` we give it a format it can handle. 

Does my explanation make sense?

ninanomenon (approved):

the-infinity:
Looks good! Could you document the findings / changes in the mapping document? Like `type is always PARKING_CENTER_LINE which is a LinearParkingSiteType` change

PR #410, opened 2026-07-16
change name to readable names from parking type

AbdullahiFatola:
This PR updates the `name` mapping to have readable names from the attribute `Parkplat-Typ`.

### DIFF
--- a/src/parkapi_sources/converters/esslingen/validator.py
+++ b/src/parkapi_sources/converters/esslingen/validator.py
@@ -117,7 +117,7 @@ class EsslingenParkingSiteFeatureInput:
 
         static_parking_site_input = StaticParkingSiteInput(
             uid=str(self.properties.fid),
-            name=str(self.properties.parking_type),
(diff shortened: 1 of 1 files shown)


AbdullahiFatola:
@the-infinity please review, we plan to have the fix in next release.

src/parkapi_sources/converters/esslingen/validator.py
```diff
@@ -117,7 +117,7 @@ def to_static_parking_site(self) -> StaticParkingSiteInput | None:
 
         static_parking_site_input = StaticParkingSiteInput(
             uid=str(self.properties.fid),
-            name=str(self.properties.fid),
+            name=str(self.properties.parking_type),
```
the-infinity: `self.properties.parking_type` can be `None`, which gives unexpected names here.

According to docs the mapping definition, it's required anyway, so the validator should not be Nonable

AbdullahiFatola: That's right. I have added a default name `Straßen-Parkplatz` if it is `None`

PR #416, opened 2026-08-02
Nagold Bike Converter

the-infinity:
Nagold Bike


AbdullahiFatola:
@the-infinity the city sent an updated dataset and the mapping documentation has been updated here https://github.com/ParkenDD/parkapi-sources-v3/pull/417 

The updated mapping needs to be considered in this converter.

the-infinity:
Updated the converter according to mapping. Also fixes some gaps in docs. Are you fine with that, @AbdullahiFatola ?

AbdullahiFatola:
> Updated the converter according to mapping. Also fixes some gaps in docs. Are you fine with that, @AbdullahiFatola 

It seems you still have the old data for the validators. The updated data already contains full attribute names and not truncated to 10 characters. That is why I updated the mapping documentation. I uploaded the updated data [here](https://github.com/mobidata-bw/ipl-planning/blob/main/data/bike_park/Stadt_Nagold/Radabstellanlagen/Fahrradabstellanlagen_Nagold_Stand_06_08_2026.geojson)

docs/mapping/nagold_bike.md
```diff
@@ -1,46 +1,73 @@
 # Nagold Bike Parking
 
-The city of Nagold publishes a GeoJSON dataset with locations of bicycle parking installations across the city. Each feature describes a group of bicycle stands or racks available for public use.
+The city of Nagold publishes a GeoJSON dataset with locations of bicycle parking installations across the city. Each
+feature describes a group of bicycle stands or racks available for public use.
 
+The dataset is exported from a shapefile-based cadastre, therefore all property names are truncated to ten characters
+(`Anzahl_Bue` for `Anzahl_Buegel_Stellplaetze`, `Immer_geoe` for `Immer_geoeffnet_zugaenglich` and so on). The tables
```
AbdullahiFatola, line 7: The updated dataset is no more truncated to 10 characters.

src/parkapi_sources/converters/nagold_bike/validator.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def to_supervision_type(self) -> SupervisionType | None:
+        return {
+            self.YES: SupervisionType.YES,
+            self.NO: SupervisionType.NO,
+        }.get(self)
+
+
+@validataclass
+class NagoldBikePropertiesInput:
```
AbdullahiFatola, line 73: The properties now contain full attribute names not truncated to 10 characters.

docs/mapping/nagold_bike.md
the-infinity: The current version in https://github.com/mobidata-bw/ipl-planning/tree/main/data/bike_park/Stadt_Nagold/Radabstellanlagen still has the 10 character truncation. Can you provide the new file you reference to?

AbdullahiFatola: I hyperlinked it in my 2nd comment above. 
The referenced data is here: https://github.com/mobidata-bw/ipl-planning/blob/main/data/bike_park/Stadt_Nagold/Radabstellanlagen/Fahrradabstellanlagen_Nagold_Stand_06_08_2026.geojson

the-infinity:
Used the updated file now.

AbdullahiFatola (approved):
LGTM

